Issue #15, opened 2015-03-16
intern keys of object state


davisagli:
We might want to intern() the keys of the persistent object state as they are updated in the instance dict. i.e.

```
    def __setstate__(self, state):
        for k, v in state.items():
            self.__dict__[intern(k)] = v
```

cPickle normally does this via the BUILD opcode (https://github.com/python/cpython/blob/master/Modules/_pickle.c#L5958), but ZODB pickles the object type and state separately, so that opcode is not used.

Experimentation on one client project shows about 25% memory savings when loading all instances of a particular commonly used class, with negligible effect on performance. But more careful evaluation is warranted.

(Memory use could also be reduced by using a class with slots, but that requires more invasive changes to the application code.)

tseaver:
SLAGITM

mgedmin:
Seems like a good idea to me too.

tseaver:
Via #16.

PR #16, opened 2015-03-22
Intern keys in `__setstate__`

davisagli:
Implementation of the feature discussed in zopefoundation/persistent#15

I'd appreciate a close review of the C implementation. It's based on code in cPickle (https://github.com/python/cpython/blob/master/Modules/_pickle.c#L5956) but I'm just pretending to know what I'm doing with pointers and with refcounting.

### DIFF
--- a/persistent/_compat.py
+++ b/persistent/_compat.py
@@ -18,6 +18,7 @@ if sys.version_info[0] > 2: #pragma NO COVER
     import copyreg as copy_reg
     from collections import UserDict as IterableUserDict
     from collections import UserList
+    from sys import intern
 
     def _u(s):
         return s
@@ -52,3 +53,5 @@ else: #pragma NO COVER
 
     PYTHON3 = False
     PYTHON2 = True
+
+    intern = intern
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -534,6 +534,8 @@ pickle___setstate__(PyObject *self, PyObject *state)
     if (state != Py_None)
     {
         PyObject **dict;
+        PyObject *d_key, *d_value;
+        Py_ssize_t i;
 
         dict = _PyObject_GetDictPtr(self);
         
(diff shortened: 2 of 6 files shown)


persistent/_compat.py
```diff
@@ -52,3 +53,5 @@ def _native(s):
 
     PYTHON3 = False
     PYTHON2 = True
+
+    intern = intern
```
mgedmin: Is this to appease pylint/pyflakes or something?  A comment might make it clearer.

mgedmin:
I'm not qualified enough to review the C bits, sorry!

persistent/_compat.py
tseaver: It makes the following line work cleanly:

```python
from persistent._compat import intern
```

persistent/cPersistence.c
```diff
@@ -552,8 +554,20 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
 
         PyDict_Clear(*dict);
-        if (PyDict_Update(*dict, state) < 0)
-            return NULL;
+
+        i = 0;
+        while (PyDict_Next(state, &i, &d_key, &d_value)) {
+            /* normally the keys for instance attributes are
+               interned.  we should try to do that here. */
+            Py_INCREF(d_key);
```
tseaver: Nitpick:  I'd rather have the `Py_INCREF` and `Py_DECREF` wrapped around the `INTERN_INPLACE` call (the thing which could actually decref the key), and remove the need for the extra decref in the failure case for `PyObject_SetItem`.

tseaver:
Except for my nitpick, LGTM.

persistent/cPersistence.c
davisagli: Still trying to wrap my head around the refcounting here...if I've got it right, I think Py_INCREF is here to convert the "borrowed" reference from PyDict_Next into an "owned" reference so that INTERN_INPLACE does not have a chance of triggering the object's finalizer. And then we no longer need to own the reference we pass to PyObject_SetItem because it will do its own incref for the reference stored in the object. Yes? If so I agree with your proposed change.

davisagli:
Updated re the nitpick.

tseaver:
Thank you very much!

PR #17, opened 2015-03-31
fix test___setstate___interns_dict_keys failure on pypy3

davisagli:
(by disabling the test, since pypy3 doesn't guarantee identity of intern'd strings)

Not using unittest.skipIf because it doesn't exist in python 2.6.

Also enable travis tests for 3.4 and pypy3


tseaver:
Thanks for the quick fix!

Issue #18, opened 2015-04-07
C and Python TimeStamp objects have different `.raw()` and `repr()`  values for same input and are not equal


jamadden:
This cropped up in testing ZODB (see zopefoundation/ZODB#33), which relies on the particular repr of known TimeStamp objects, but it's not hard to imagine this having implications in many other ways. 

Here's a doctest-like example:

First, verify we have access to a C implementation:

```python
>>> import sys
>>> sys.version
'2.7.9 (default, Dec 13 2014, 15:13:49) \n[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.56)]'
>>> import persistent.TimeStamp
>>> persistent.TimeStamp.TimeStamp
<built-in function TimeStamp>
```

Next, set up a constant time (this exact floating point value and timestamp come from ZODB tests):

```python
>>> import time
>>> now = 1229959248.3
>>> val = (time.gmtime(now)[:5] + (now % 60,))
>>> val
(2008, 12, 22, 15, 20, 48.299999952316284)
```

Create a C and Python TimeStamp:

```python
>>> C_ts = persistent.TimeStamp.TimeStamp(*val)
>>> Py_ts = persistent.TimeStamp.pyTimeStamp(*val)
```

Not only do they have different reprs and raw values (different in the last character):

```
>>> print(repr(C_ts))
'\x03z\xbd\xd8\xce\x14z\xdd'
>>> print(repr(Py_ts))
'\x03z\xbd\xd8\xce\x14z\xde'
>>> C_ts.raw()
'\x03z\xbd\xd8\xce\x14z\xdd'
>>> Py_ts.raw()
'\x03z\xbd\xd8\xce\x14z\xde'
```

They're not even equal to each other, despite having the same input:

```python
>>> C_ts == Py_ts
False
```

To me this seems like a bug. I can try to put together a PR, but I'm not familiar with the code so it might take me awhile.

jamadden:
I've got this now, preparing a PR.

PR #19, opened 2015-04-07
Make the .raw() and repr() of the Python TimeStamp match the C version in more cases.

jamadden:
It turns out that the Python version was rounding the seconds where the C version was doing nothing so complicated.

Also, make the behaviour of comparisons and hashing match as well. This should all be tested.

Fixes zopefoundation/persistent#18. Relates to zopefoundation/ZODB#33 and fixes at least some of the issues there (there will probably be more PRs coming for the rest).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,11 @@
 
 - 100% branch coverage.
 
+- Make the C and Python `TimeStamp` objects behave more alike. The
+  Python version now produces the same `repr` and `.raw()` output as
+  the C version, and has the same hashcode. In addition, the Python
+  version is now supports ordering and equality like the C version.
+
 4.0.8 (2014-03-20)
 ------------------
 
@@ -18,7 +23,7 @@
 
 - In pure-Python ``Persistent``, avoid loading state in ``_p_activate``
   for non-ghost objects (which could corrupt their state).  (PR #9)
-    
+
 - In pure-Python, and don't throw ``POSKeyError`` if ``_p_activate`` is
   called on an object that has never been committed.  (PR #9)
 
--- a/persistent/tests/test_timestamp.py
+++ b/persistent/tests/test_timestamp.py
@@ -11,6 +11,7 @@
 # FOR A PARTICULAR PURPOSE.
 #
 ##############################################################################
+import operator
 import unittest
 
 class Test__UTC(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


jamadden:
Whoops, turns out ZODB tests also depend on the `__str__` value too. I'll add that to this PR as well.

CHANGES.rst
```diff
@@ -11,6 +11,11 @@
 
 - 100% branch coverage.
 
+- Make the C and Python `TimeStamp` objects behave more alike. The
+  Python version now produces the same `repr` and `.raw()` output as
+  the C version, and has the same hashcode. In addition, the Python
+  version is now supports ordering and equality like the C version.
+
```
mgedmin: Nitpick: I believe the convention is two blank lines between sections.

(It probably doesn't matter.  Why am I even commenting?)

mgedmin:
LGTM, mostly.

I'd be tempted to get rid of several of those '# pragma: nocover' comments and instead add tests comparing PyTimeStamps with something else.

Am I right in assuming Python 2 versus 3 makes that tricky?

CHANGES.rst
jamadden: Ahh. It's just a bullet point, not a section, and the rest of the file has single lines between bullet points just like this. (It is somewhat inconsistent on the use of single vs double backticks, though, I went with what seemed newest.)

jamadden:
@mgedmin Yes, Python 3's behaviour of raising `TypeError` when you return `NotImplemented` make it tricky to cover those lines in comparison to Python 2's simply returning `False` (not that it can't be done, it just needs separate test cases that didn't really seem to add much value to me.) I suppose we could call the dunder methods directly, but that nearly defeats the point of making sure the operator methods are actually correctly named and implemented to be called by the interpreter.

CHANGES.rst
mgedmin: Sorry, I was unclear.  What I really meant was "two blank lines above the 4.0.8 version heading".

(This rule is not maintained consistently: e.g. there's only one blank line above 4.0.7, but two blank lines above 4.0.6, 4.0.5, and 4.0.4.  There *were* two blank lines above the 4.0.8 heading before your changes, which is why I commented.)

tseaver:
@jamadden As you've found elsewhere, `#pragma NO COVER` is where the bugs run to hide when you flip the light switch on.  Maybe the PR could define a straddling helper function / method which dealt with the `return False` vs. `raise TypeError` bits.

jamadden:
Sure, no problem, added.

tseaver:
Thanks!

Issue #21, opened 2015-04-09
Test failure in test_hash_equal on py2.6 windows 


agroszer:
Commit 2a8313ccad80cbc01f5153a55788ab8b5412bc04 breaks tests on winbot

see here:

http://winbot.zope.org/builders/persistent_py_265_32/builds/971/steps/test/logs/stdio

agroszer:
@jamadden ^

mgedmin:
I have a feeling this is because the tests run on a 32-bit build rather than because it's Windows or because it's Python 2.6.

Note that `hash('a string')` will differ on 32- and 64-bit platforms (and also different Python versions may compute different hashes):

```python
>>> hash('a string')  # Python 2.6.5, Linux, 32-bit
1827144452
>>> hash('a string')  # Python 2.6.9, Linux, 64-bit
-8411828025894108412
>>> hash('a string')  # Python 3.4.2, Linux, 64-bit
-8227545103199816438
```

mgedmin:
At some point Travis CI might allow us to test both 32- and 64-bit builds in the build matrix: https://github.com/travis-ci/travis-ci/issues/986

mgedmin:
Okay, string hashes are irrelevant for this issue: https://github.com/zopefoundation/persistent/blob/663497de550ef21c561480bfc49b7d4449dcdd04/persistent/_timestamp.c#L155-L167

mgedmin:
Perhaps it would be sufficient to replace `ctypes.c_int64` with `ctypes.c_long` in pyTimeStamp.\__hash__?

jamadden:
It will depend on the platform. The C implementation uses `long`, which can be either 4 or 8 bytes. It's [always 4 bytes](https://msdn.microsoft.com/en-us/library/cc953fe1.aspx) under Microsoft compilers, [I think](http://ivbel.blogspot.com/2012/02/size-of-primitive-types-in-c-language.html), but under gcc/clang it depends on whether the code is being compiled as 64-bit or not. 

Rather than trying to detect those differences in python, what about making the C implementation use `long long`, which in practice is always 8 bytes? That way the hash is consistent regardless of platform, and the Python code doesn't have to try to figure it out.

jamadden:
Oh, nuts, that won't work, the signature of the C type object `hashfunc` is incompatible. Never mind.

mgedmin:
[...]

PR #23, opened 2015-04-10
Fix tests for TimeStamp hashcode under 32-bit Pythons.

jamadden:
Fixes zopefoundation/persistent#21 again. This time I found a 32-bit version of Python 2.6 to test on so the fix should be good.


tseaver:
Thanks for the quick follow-up.

PR #31, opened 2015-12-23
Get rid of ctypes dependency to make persistent more pure-python compatible

michwill:
In order to work with max integers, persistent uses ctypes. It is not really necessary but breaks compatibility with pypy.js, for example.

### DIFF
--- a/persistent/tests/test_timestamp.py
+++ b/persistent/tests/test_timestamp.py
@@ -268,16 +268,14 @@ class PyAndCComparisonTests(unittest.TestCase):
         # platform and test the same; also verify 64 bit
         bit_32_hash = -1419374591
         bit_64_hash = -3850693964765720575
-        import persistent.timestamp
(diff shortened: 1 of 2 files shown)


tseaver:
Breaks Py3k, which has no `sys.maxint`.

michwill:
Indeed. Is it correct to use sys.maxsize for Py3k?

tseaver:
@michwill `sys.maxsize` should work on all the platforms we care about.

tseaver:
Rather than  monkey-patching `sys`, please initialize a constant (e.g., `_MAXSIZE`) in the `timestamp` module, and then monkey-patch *that* value in the tests.

tseaver:
I will tackle fixing up the branch to use `sys.maxsize` and do the monkey-patch I suggested.

tseaver:
@michwill Thank you for the patch!

PR #34, opened 2016-04-14
Don't use/poison global wheel cache when building w/ PURE_PYTHON.

tseaver:
The `none-any` wheels built in the PURE_PYTHON environments should not poison the user's global wheel cache.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@
 4.2.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


jamadden:
That's a good point I hadn't ever considered. 

This isn't really tested on Travis, but it does what it says locally for me.

tseaver:
The ugliest thing about the poisoning is that *other* projects tests can begin to fail after running the `-pure` ones here (or in `persistent`, `BTrees`, etc.).

PR #37, opened 2016-05-05
TimeStamp: sub-second precision for Python implementation of timeTime()

jmuchemb:
The fractional part of return values was always 0.

### DIFF
--- a/persistent/tests/test_timestamp.py
+++ b/persistent/tests/test_timestamp.py
@@ -100,9 +100,10 @@ class pyTimeStampTests(unittest.TestCase):
         self.assertEqual(ts.timeTime(), DELTA_SECS)
 
     def test_ctor_from_string_non_zero(self):
-        before = self._makeOne(2011, 2, 16, 14, 37, 22.0)
+        before = self._makeOne(2011, 2, 16, 14, 37, 22.80544)
         after = self._makeOne(before.raw())
         self.assertEqual(before.raw(), after.raw())
+        self.assertEqual(before.timeTime(), 1297867042.80544)
 
     def test_ctor_from_elements(self):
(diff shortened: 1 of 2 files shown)


persistent/tests/test_timestamp.py
```diff
@@ -100,9 +100,10 @@ def test_ctor_from_string(self):
         self.assertEqual(ts.timeTime(), DELTA_SECS)
 
     def test_ctor_from_string_non_zero(self):
-        before = self._makeOne(2011, 2, 16, 14, 37, 22.0)
+        before = self._makeOne(2011, 2, 16, 14, 37, 22.80544)
         after = self._makeOne(before.raw())
         self.assertEqual(before.raw(), after.raw())
+        self.assertEqual(before.timeTime(), 1297867042.80544)
```
tseaver: Magic numbers (which can't be verified by eye) are problematic:  could we have generated the timestamp using the `time` module and the constants above?

jmuchemb: There's `time.mktime` but it's for local times. On my machine (GMT+1):
```python
>>> time.mktime((2011, 2, 16, 14, 37, 22, 0, 0, 0)) - 1297867042
3600
```
(with dummy values for `tm_wday` and `tm_yday`)

I don't find such expected values problematic, when the alternative is to duplicate the tested code. And in case where hardcoding values in the code is unmaintainable, I prefer a mechanism with « dumps » (the idea is to have a mode in which the test suite generates dumps instead of comparing values against them). I have started to do for a big test suite (45k lines excluding dumps) because I was tired of implementing many features twice.

```
# date -u -d @1297867042
Wed Feb 16 14:37:22 UTC 2011
```

PR #39, opened 2016-05-26
Fix #38 by returning the correct type from TimeStamp_hash.

### LINKED ISSUE #38: Tests fails on Windows, Python3, 64bit machine, likely due to float == double on windows
The failures are

```
FAIL: test_hash_equal (persistent.tests.test_timestamp.PyAndCComparisonTests)
  File "c:\projects\persistent\persistent\tests\test_timestamp.py", line 260, in test_hash_equal
    self.assertEqual(hash(c), hash(py))
AssertionError: 2875592705 != -1419374591

FAIL: test_hash_equal_constants (persistent.tests.test_timestamp.PyAndCComparisonTests)
  File "c:\projects\persistent\persistent\tests\test_timestamp.py", line 327, in test_hash_equal_constants
    self.assertEqual(hash(c), -721379967)
AssertionError: 3573587329 != -721379967
```

The hashes are equal mod 2**32. The tests pass on winbot, which is running a 32-bit Python.

The failing AppVeyor build is using a 64-bit Python.

This looks like a genuine bug in the test suite:

```
    def test_hash_equal_constants(self):
        ...
        is_32_bit = persistent.timestamp.c_long == ctypes.c_int32 or _is_jython                                                                     
        ....
        if is_32_bit:
            self.assertEqual(hash(c), -721379967)
        else:
            self.assertEqual(hash(c), 1000006000001)
```

On 64-bit Windows long is 32-bits long.

jamadden:


persistent/_timestamp.c
```diff
@@ -152,7 +152,12 @@ TimeStamp_richcompare(TimeStamp *self, TimeStamp *other, int op)
     return result;
 }
 
+
+#ifdef PY3K
+static Py_hash_t
+#else
 static long
+#endif
```
tseaver: Nice catch!  Logically, this kind of switch would be better as a `#define` in https://github.com/zopefoundation/persistent/blob/master/persistent/_compat.h.

jamadden: I thought about that, but (1) it's only used once and (2) more irkingly I couldn't decide on what a good name would be.

PR #44, opened 2016-07-21
On deactivate release in-slots objects too

navytux:
On ._p_deactivate() and ._p_invalidate(), when an object goes to ghost
state, objects referenced by all its attributes, except related to
persistence machinery, are released, this way freeing memory (if they
were referenced only from going-to-ghost object).

That's the idea - an object in ghost state is simply a stub, which loads
its content on first access (via hooking into get/set attr) while
occupying minimal memory in not-yet-loaded state.

However the above is not completely true right now, as currently on
ghostification only object's .__dict__ is released, while in-slots objects
are retained attached to ghost object staying in RAM:

    ---- 8< ----
    from ZODB import DB
    import gc

    db = DB(None)
    jar = db.open()

    class C:
        def __init__(self, v):
            self.v = v
        def __del__(self):
            print 'released (%s)' % self.v

    class P1(Persistent):
        pass

    class P2(Persistent):
        __slots__ = ('aaa')

    p1 = P1()
    jar.add(p1)
    p1.aaa = C(1)

    p2 = P2()
    jar.add(p2)
    p2.aaa = C(2)

    p1._p_invalidate()
    # "released (1)" is printed

    p2._p_invalidate()
    gc.collect()
    # "released (2)" is NOT printed     <--
    ---- 8< ----

So teach ghostify() & friends to release objects in slots to free-up
memory when an object goes to ghost state.

NOTE PyErr_Occurred() added after ghostify() calls because
pickle_slotnames() can raise an error, but we do not want to change
ghostify() prototype for backward compatibility reason - as it is used
in cPersistenceCAPIstruct.

( I hit this bug with wendelin.core which uses proxies to load
  data from DB to virtual memory manager and then deactivate proxy right
  after load has been completed:

  https://lab.nexedi.com/nexedi/wendelin.core/blob/f7803634/bigfile/file_zodb.py#L239
  https://lab.nexedi.com/nexedi/wendelin.core/blob/f7803634/bigfile/file_zodb.py#L295 )

/cc @jimfulton, @jmuchemb

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -82,6 +82,7 @@ fatal_1350(cPersistentObject *self, const char *caller, const char *detail)
 #endif
 
 static void ghostify(cPersistentObject*);
+static PyObject * pickle_slotnames(PyTypeObject *cls);
 
 static PyObject * convert_name(PyObject *name);
 
@@ -150,7 +151,7 @@ accessed(cPersistentObject *self)
 static void
 ghostify(cPersistentObject *self)
 {
-    PyObject **dictptr;
+    PyObject **dictptr, *slotnames;
 
     /* are we already a ghost? */
     if (self->state == cPersistent_GHOST_STATE)
@@ -180,12 +181,56 @@ ghostify(cPersistentObject *self)
         _estimated_size_in_bytes(self->estimated_size);
     ring_del(&self->ring);
     self->state = cPersistent_GHOST_STATE;
+
+    /* clear __dict__ */
     dictptr = _PyObject_GetDictPtr((PyObject *)self);
     if (dictptr && *dictptr)
     {
         Py_DECREF(*dictptr);
         *dictptr = NULL;
     }
+
+    /* clear all slots besides _p_* */
+    slotnames = pickle_slotnames(Py_TYPE(self));
+    if (slotnames && slotnames != Py_None)
+    {
+        int i;
+
+        for (i = 0; i < PyList_GET_SIZE(slotnames); i++)
+        {
+            PyObject *name;
+            char *cname;
+            int is_special;
+
+            name = PyList_GET_ITEM(slotnames, i);
+#ifdef PY3K
+            if (PyUnicode_Check(name))
+            {
+                PyObject *converted = convert_name(name);
+                cname = PyBytes_AS_STRING(converted);
+#else
+            if (PyBytes_Check(name))
+            {
+                cname = PyBytes_AS_STRING(name);
+#endif
+                is_special = !strncmp(cname, "_p_", 3);
+#ifdef PY3K
+                Py_DECREF(converted);
+#endif
+                if (is_special) /* skip persistent */
+                {
+                    continue;
+                }
+            }
+
+            /* NOTE: this skips our delattr hook */
+            if (PyObject_GenericSetAttr((PyObject *)self, name, NULL) < 0)
+                /* delattr of non-set slot will raise AttributeError - we
+                 * simply ignore. */
+                PyErr_Clear();
+        }
+    }
+    Py_XDECREF(slotnames);
 
     /* We remove the reference to the just ghosted object that the ring
     * holds.  Note that the dictionary of oids->objects has an uncounted
@@ -270,6 +315,8 @@ Per__p_deactivate(cPersistentObject *self)
             called directly. Methods that override this need to
             do the same! */
         ghostify(self);
+        if (PyErr_Occurred())
+            return NULL;
     }
 
     Py_INCREF(Py_None);
@@ -298,6 +345,8 @@ Per__p_invalidate(cPersistentObject *self)
         if (Per_set_changed(self, NULL) < 0)
             return NULL;
         ghostify(self);
+        if (PyErr_Occurred())
+            return NULL;
     }
     Py_INCREF(Py_None);
     return Py_None;
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -1319,6 +1319,27 @@ class _Persistent_Base(object):
         self.assertEqual(list(jar._loaded), [])
         self.assertEqual(list(jar._registered), [])
 
+    def test__p_invalidate_from_changed_w_slots(self):
+        class Derived(self._getTargetClass()):
+            __slots__ = ('myattr1', 'myattr2')
+            def __init__(self):
+                self.myattr1 = 'value1'
+                self.myattr2 = 'value2'
+        inst, jar, OID = self._makeOneWithJar(Derived)
+        inst._p_activate()
+        inst._p_changed = True
+        jar._loaded = []
+        jar._registered = []
+        self.assertEqual(Derived.myattr1.__get__(inst), 'value1')
+        self.assertEqual(Derived.myattr2.__get__(inst), 'value2')
+        inst._p_invalidate()
+        self.assertEqual(inst._p_status, 'ghost')
+        self.assertEqual(list(jar._loaded), [])
+        self.assertRaises(AttributeError, lambda: Derived.myattr1.__get__(inst))
+        self.assertRaises(AttributeError, lambda: Derived.myattr2.__get__(inst))
+        self.assertEqual(list(jar._loaded), [])
+        self.assertEqual(list(jar._registered), [])
+
     def test__p_invalidate_from_sticky(self):
         inst, jar, OID = self._makeOneWithJar()
         inst._p_activate() # XXX
--- a/persistent/persistence.py
+++ b/persistent/persistence.py
@@ -303,11 +303,11 @@ class Persistent(object):
                     _OGA(self, '_p_register')()
         object.__delattr__(self, name)
 
-    def _slotnames(self):
+    def _slotnames(self, _v_exclude=True):
         slotnames = copy_reg._slotnames(type(self))
         return [x for x in slotnames
                    if not x.startswith('_p_') and
-                      not x.startswith('_v_') and
+                      not (x.startswith('_v_') and _v_exclude) and
                       not x.startswith('_Persistent__') and
                       x not in Persistent.__slots__]
 
@@ -423,6 +423,10 @@ class Persistent(object):
         idict = getattr(self, '__dict__', None)
         if idict is not None:
             idict.clear()
+        type_ = type(self)
+        for slotname in Persistent._slotnames(self, _v_exclude=False):
+            getattr(type_, slotname).__delete__(self)
+
         # Implementation detail: deactivating/invalidating
         # updates the size of the cache (if we have one)
         # by telling it this object no longer takes any bytes


persistent/cPersistence.c
```diff
@@ -180,13 +181,57 @@ ghostify(cPersistentObject *self)
...
+#else
+            if (PyBytes_Check(name))
+            {
+                cname = PyBytes_AS_STRING(name);
+#endif
+                is_special = !strncmp(cname, "_p_", 3);
+#ifdef PY3K
+                Py_DECREF(converted);
+#endif
+                if (is_special) /* skip persistent */
```
tseaver: Better to make the test `if ( ! is_special)` and then clear the attribute inside, rather than `continue` for the other case.

persistent/cPersistence.c
```diff
@@ -180,13 +181,57 @@ ghostify(cPersistentObject *self)
...
+                {
+                    continue;
+                }
+            }
+
+            /* NOTE: this skips our delattr hook */
+            if (PyObject_GenericSetAttr((PyObject *)self, name, NULL) < 0)
```
tseaver: Unless my eyes are fooling me, because it is outside the look, this only clears the *last* slot's value.

persistent/tests/test_persistence.py
```diff
@@ -1319,6 +1319,23 @@ def __init__(self):
         self.assertEqual(list(jar._loaded), [])
         self.assertEqual(list(jar._registered), [])
 
+    def test__p_invalidate_from_changed_w_slots(self):
+        class Derived(self._getTargetClass()):
+            __slots__ = ('myattr')
```
tseaver: Try this with mutilple slots to exercise the bug I'm describing above.

navytux:
@tseaver thanks for feedback. I've rechecked - setattr call is at the right
place, and to make sure it is indeed so, I've updated the patch and added
another attribute to test, as you adviced. Interdiff:

```diff
diff --git a/persistent/tests/test_persistence.py b/persistent/tests/test_persistence.py
index f82d42d..170b337 100644
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -1321,19 +1321,23 @@ class _Persistent_Base(object):

     def test__p_invalidate_from_changed_w_slots(self):
         class Derived(self._getTargetClass()):
-            __slots__ = ('myattr')
+            __slots__ = ('myattr1', 'myattr2')
             def __init__(self):
-                self.myattr = 'value'
+                self.myattr1 = 'value1'
+                self.myattr2 = 'value2'
         inst, jar, OID = self._makeOneWithJar(Derived)
         inst._p_activate()
         inst._p_changed = True
         jar._loaded = []
         jar._registered = []
+        self.assertEqual(Derived.myattr1.__get__(inst), 'value1')
+        self.assertEqual(Derived.myattr2.__get__(inst), 'value2')
         inst._p_invalidate()
         self.assertEqual(inst._p_status, 'ghost')
         self.assertEqual(list(jar._loaded), [])
-        self.assertEqual(hasattr(inst, 'myattr'), False)
-        self.assertEqual(list(jar._loaded), [OID])  # hasattr makes it loaded
+        self.assertRaises(AttributeError, lambda: Derived.myattr1.__get__(inst))
+        self.assertRaises(AttributeError, lambda: Derived.myattr2.__get__(inst))
+        self.assertEqual(list(jar._loaded), [])
         self.assertEqual(list(jar._registered), [])

     def test__p_invalidate_from_sticky(self):
```

the `is_special` handling was copied (and thus kept uniform) from pickle_copydict():

https://github.com/zopefoundation/persistent/blob/b29c0ca2/persistent/cPersistence.c#L371

and from pickle__getstate__():

https://github.com/zopefoundation/persistent/blob/b29c0ca2/persistent/cPersistence.c#L453

To me it makes sense to keep code uniform to itself, so I do not change it for now.

Please let me know whether you insist on this.

Thanks again,
Kirill

navytux:
@tseaver ping. Maybe you overlooked my last reply. I've adjusted the patch to test for several in-slot attributes there. Thanks beforehand for feedback. Kirill

navytux:
@tseaver maybe you overlooked my reply from 3 weeks ago: the patch was corrected and tests are there to cover the case you've asked about. To me there is no reason not to apply this patch, so please, if there is anything wrong or not right on my side, please provide feedback.

And in general I would appreciate maintainers feedback about the patch and/or applying it.

Thanks beforehand again,  
Kirill

/cc @jimfulton

tseaver:
@navytux Thanks for your efforts!

navytux:
@tseaver thanks for applying the patch.

jamadden:
With this change, the ZODB tests are broken, at least for PyPy. There are errors like this:
```
  File "//pypy/lib-python/2.7/unittest/case.py", line 331, in run
    testMethod()
  File "//ZODB/src/ZODB/scripts/tests/test_repozo.py", line 815, in test_w_changes
    db.mutate()
  File "//ZODB/src/ZODB/scripts/tests/test_repozo.py", line 84, in mutate
    tree[random.randrange(100000)] = random.randrange(100000)
  File "//site-packages/BTrees/_base.py", line 812, in __setitem__
    self._set(self._to_key(key), self._to_value(value))
  File "//persistent/persistent/persistence.py", line 266, in __getattribute__
    oga(self, '_p_activate')()
  File "//persistent/persistent/persistence.py", line 386, in _p_activate
    jar.setstate(self)
  File "//ZODB/src/ZODB/Connection.py", line 771, in setstate
    self._reader.setGhostState(obj, p)
  File "//ZODB/src/ZODB/serialize.py", line 623, in setGhostState
    obj.__setstate__(state)
  File "//site-packages/BTrees/_base.py", line 1065, in __setstate__
    self.clear()
  File "//site-packages/BTrees/_base.py", line 818, in clear
    if self._data:
  File "//persistent/persistent/persistence.py", line 268, in __getattribute__
    return oga(self, name)
AttributeError: _data
```

(BTrees do keep `_data` in `__slots__`.)

jamadden:
BTrees set `_data` in `__new__` and expect for that attribute to stay present. It wouldn't surprise me to find this pattern elsewhere. So this seems like a breaking backwards incompatible change. If it's going to stay then I think persistent's major version number needs to go up.

navytux:
@jamadden thanks for feedback and heads up. To me it was never a promise from ZODB to keep slotted objects alive during ghostification and from this point of view we should better fix it all there. I've made 2 patches:

https://github.com/zopefoundation/persistent/pull/48
https://github.com/zopefoundation/BTrees/pull/47

with them and PURE_PYTHON=1 whole BTrees testsuite, including its ZODB part, is now passing fully again.

About ZODB itself: I get the following testsuite failures on today's ZODB master (e0a0ec36) with persistent, BTrees and transaction also being taken from their latest masters + my 2 patches:

```
FAIL: doctest_invalidate (ZODB.tests.testConnection.InvalidationTests)
Doctest: ZODB.tests.testConnection.InvalidationTests.doctest_invalidate
  File "/usr/lib/python2.7/doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for ZODB.tests.testConnection.InvalidationTests.doctest_invalidate
  File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/tests/testConnection.py", line 513, in doctest_invalidate

File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/tests/testConnection.py", line 529, in ZODB.tests.testConnection.InvalidationTests.doctest_invalidate
Failed example:
    transaction.commit()
Exception raised:
      File "/usr/lib/python2.7/doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest ZODB.tests.testConnection.InvalidationTests.doctest_invalidate[9]>", line 1, in <module>
        transaction.commit()
      File "/home/kirr/src/wendelin/z/transaction/transaction/_manager.py", line 123, in commit
        return self.get().commit()
      File "/home/kirr/src/wendelin/z/transaction/transaction/_transaction.py", line 280, in commit
        reraise(t, v, tb)
      File "/home/kirr/src/wendelin/z/transaction/transaction/_transaction.py", line 271, in commit
        self._commitResources()
      File "/home/kirr/src/wendelin/z/transaction/transaction/_transaction.py", line 416, in _commitResources
        reraise(t, v, tb)
      File "/home/kirr/src/wendelin/z/transaction/transaction/_transaction.py", line 390, in _commitResources
        rm.commit(self)
      File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/Connection.py", line 481, in commit
        self._commit(transaction)
      File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/Connection.py", line 530, in _commit
        self._store_objects(ObjectWriter(obj), transaction)
      File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/Connection.py", line 593, in _store_objects
        self._cache[oid] = obj
      File "/home/kirr/src/wendelin/z/t.venv/local/lib/python2.7/site-packages/persistent/picklecache.py", line 161, in __setitem__
        self.data[oid] = value
      File "/usr/lib/python2.7/weakref.py", line 108, in __setitem__
        self.data[key] = KeyedRef(value, self._remove, key)
      File "/usr/lib/python2.7/weakref.py", line 278, in __new__
        self = ref.__new__(type, ob, callback)
    TypeError: cannot create weak reference to 'Persistent' object
File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/tests/testConnection.py", line 563, in ZODB.tests.testConnection.InvalidationTests.doctest_invalidate
Failed example:
    p2._p_state
Expected:
    -1
Got:
    0


FAIL: testBogusObject (ZODB.tests.testCache.CacheErrors)
  File "/home/kirr/src/wendelin/z/ZODB/src/ZODB/tests/testCache.py", line 381, in testBogusObject
    self.assertEqual(rc(None), nones)
AssertionError: 104777 != 104775

Ran 1153 tests in 224.940s

FAILED (failures=2)
```

but I get the same + even more failures if I try ZODB testsuite from same ZODB but with persistent, BTrees and transaction releases (and latest persistent release is 4.2.1 which does not yet have #44 in it).

I have not tried with PyPy (only with CPython and PURE_PYTHON=1) but I think my patches should be ok to go in.

Thanks once again,
Kirill

/cc @jimfulton

jamadden:
Promise or not, there is a *very* long history of it not being done. And we have evidence that the old behavior was relied on, and changing it breaks things in prominent projects. Hence the suggestion that this should not be in a minor bug-fix release.

navytux:
I see. Any other concrete failures to look at except those which I've fixed?

jimfulton:
I'm only responding at a high level. I haven't looked closely.

+1 for this fix. Ghosts really should release their state.

+1 for a major version bump.  Slots are in the same category for me as overriding __getstate__/__setstate__.  Too clever for common use, but probably justified in some cases, such as BTrees.  I'm optimistic that slots aren't widely used and that this won't break too much. I think a version bump should be enough to keep us honest. :)

jamadden:
> I see. Any other concrete failures to look at except those which I've fixed?

I (obviously) haven't surveyed the universe of code, but one zopefoundation class I can think of that is Persistent and uses slots is [zope.container's Proxy](https://github.com/zopefoundation/zope.container/blob/master/src/zope/container/_proxy.py#L28). 

This change breaks a number of tests there. There are both `RuntimeError: maximum recursion depth exceeded` and `AttributeErrors`.

navytux:
Thanks for feedback. Indeed we can do a major version bump then and fix those cases one by one.

jamadden:
This change needs some adjustment in the way it destroys slots (and a test case!). Anything that uses `__getnewargs__` and a slot is broken by it. To be fair, anything that used `__getnewargs__` and its `__dict__` would also have been broken before this change, but the major use case of `__getnewargs__` is for immutable objects, which usually use slots. Still, that also needs a test case. 

As mentioned, a notable breakage is the zope.container ContainedProxy. As the code here stands, I don't think it's possible to fix ContainedProxy (without some extremely undesirable coupling). Code here needs to change. Here's a walkthrough.

Here's a test that fails (test_contained_zodb:test_basic_persistent_w_non_persistent_proxied):
```
    >>> p = ContainedProxy([1])
    >>> p.__parent__ = 2
    >>> p.__name__ = 'test'
    >>> db = DB(DemoStorage('test_storage'))
    >>> c = db.open()
    >>> c.root()['p'] = p
    >>> transaction.commit()

    >>> c2 = db.open()
    >>> p2 = c2.root()['p']
    >>> p2
```

Printing the `repr` of p2 (which prints `repr(self._wrapped)`) blows up because there is no `_wrapped` attribute.  

`_wrapped` is a slot that's set by `__new__`. The `ContainedProxy` object implements `__getnewargs__` and `__reduce__` to ensure that `_wrapped` is passed to `__new__`. It has to do this because that's what the base class expects. It also implements `__setstate__` to reanimate its own custom state (which is fixed to be what the C implementation provides, presumably):

```python
    def __new__(cls, wrapped):
        inst = super(PyContainedProxyBase, cls).__new__(cls, wrapped)
        inst.__parent__ = None
        inst.__name__ = None
        return inst
    def __reduce__(self):
        return (type(self),
                (self._wrapped,),
                (self.__parent__, self.__name__))
    def __setstate__(self, state):
        object.__setattr__(self, '__parent__', state[0])
        object.__setattr__(self, '__name__', state[1])
```

So far so good.

Ok, here's where things go wrong. The line `p2 = c2.root()['p']` instantiates the class and passes the newargs to it as expected:
```
      File "<doctest zope.container.tests.test_contained_zodb.test_basic_persistent_w_non_persistent_proxied[8]>", line 1, in <module>
        p2 = c2.root()['p']
      File "//lib-python/2.7/UserDict.py", line 19, in __getitem__
        if key in self.data:
      File "//persistent/persistent/persistence.py", line 266, in __getattribute__
        oga(self, '_p_activate')()
      File "//persistent/persistent/persistence.py", line 386, in _p_activate
        jar.setstate(self)
      File "//ZODB/src/ZODB/Connection.py", line 771, in setstate
        self._reader.setGhostState(obj, p)
      File "//ZODB/src/ZODB/serialize.py", line 622, in setGhostState
        state = self.getState(pickle)
      File "//ZODB/src/ZODB/serialize.py", line 615, in getState
        return unpickler.load()
      File "//site-packages/zodbpickle/pickle_2.py", line 881, in load
        dispatch[key](self)
      File "//site-packages/zodbpickle/pickle_2.py", line 938, in load_binpersid
        self.append(self.persistent_load(pid))
      File "//ZODB/src/ZODB/serialize.py", line 483, in _persistent_load
        return self.load_oid(reference)
      File "//ZODB/src/ZODB/serialize.py", line 565, in load_oid
        return self._conn.get(oid)
      File "//ZODB/src/ZODB/Connection.py", line 244, in get
        obj = self._reader.getGhost(p)
      File "//ZODB/src/ZODB/serialize.py", line 609, in getGhost
        return klass.__new__(klass, *args)
      File "//zope.container/src/zope/container/_proxy.py", line 40, in __new__
```

However, the next step in `Connection.get` is to pass the newly-created ghost to the picklecache 😉 , which in turn calls `_p_invalidate_deactivate_helper` to ensure that the object is really a ghost. This function is where the logic was added to destroy the slots:

```
      File "//ZODB/src/ZODB/serialize.py", line 565, in load_oid
        return self._conn.get(oid)
      File "//ZODB/src/ZODB/Connection.py", line 251, in get
        self._cache.new_ghost(oid, obj)
      File "//persistent/persistent/picklecache.py", line 249, in new_ghost
        obj._p_invalidate_deactivate_helper()
```

```python
    def _p_invalidate_deactivate_helper(self):
        jar = _OGA(self, '_Persistent__jar')
        ...
        idict = getattr(self, '__dict__', None)
        if idict is not None:
            idict.clear()
        type_ = type(self)
        for slotname in Persistent._slotnames(self, _v_exclude=False):
            try:
                getattr(type_, slotname).__delete__(self)
            except AttributeError:
                pass
```

Thus, when we activate the ghost, we wind up with a broken object. There's nothing the persistent object can do about this: the information it needed was only recorded as part of `__getnewargs__` and passed to `__new__`. It's not part of the information passed to `__getstate__`.

The tentative conclusion is that the way in which the picklecache interacts with `_p_invalidate_deactivate_helper` needs to change. It's actually not entirely clear to me why that method needs to do all the work that it's doing, when it's called from `PickleCache.new_ghost`, which in turn is *only* called from `Connection.get` and passed a new ghost (from the reader.getGhost).

This doesn't happen in the C implementation because `ghostify` is only called from `_p_deactivate` and doesn't interact with the picklecache and `new_ghost` at all. The cPickleCache simply directly sets the object's state to GHOST.

jamadden:
I took a stab at fixing this in issue #49 and the branch https://github.com/zopefoundation/persistent/compare/new_ghost_doesnt_clear_newargs 

Unfortunately, it doesn't seem to be fixable in pure python at this level.

In the case of a proxy, in C, the wrapped object is not available in the `__dict__` or `__slots__` *at all*, it's just in the struct, so it never gets cleared. In Python, we have no choice but to store it somewhere (and a slot is the best choice for it). But if it's only set by `__new__` then once `_p_deactivate` has been called, the object cannot be properly unghosted anymore.

Here's that same test case from above, with some inline notes. My branch gets it farther, but once we start clearing slots, it still fails:
```python
 >>> p = ContainedProxy([1])
 >>> p.__parent__ = 2
 >>> p.__name__ = 'test'
 >>> db = DB(DemoStorage('test_storage'))
 >>> c = db.open()
 >>> c.root()['p'] = p
 >>> transaction.commit()

 >>> c2 = db.open()
 >>> p2 = c2.root()['p']
 >>> p2
 [1] # Previously, we blew up here because _wrapped was gone. My branch fixes this.
 >>> p2.__parent__
 2
 >>> p2.__name__
 'test'

 >>> p2._p_changed
 0
 >>> p2._p_deactivate()
 >>> bool(p2._p_changed)
 False
 >>> p2.__name__
 'test'
 >>> p2
 [1] # Now we blow up here. Once _p_deactivate has been called, we have no way to reconstitute _wrapped
```

Here are our options as I see them:

1. Stick with the status quo: don't clear `__slots__`.
2. Require every subclass who uses `__new__` and `__reduce__` or `__getnewargs__` to *also* start overriding `_p_deactivate` and only clear things that they can reconstitute from `__setstate__`.

Breaking pickle compatibility might be an option in some cases to store more information in the value passed to `__setstate__`, but it's not an appealing one.

Using `__new__` and `__reduce__`/`__getnewargs__` are relatively advanced things to do, but they're within the realm of standard Python and standard pickle. People can be expected to be using them. But overriding `_p_deactivate` is persistent-specific and even more advanced (and discouraged?).

Have I missed anything? 

My vote at this point is leaning towards the status quo and not clearing `__slots__`.

navytux:
@jamadden please forgive me in advance - I'm not so familiar with pure-python persistent details (we only use C implementation), but is maybe the following fact help this case: in pure-python case from slots that are cleared `_p_*` and `_Persistent__*` names are ommitted. I guess `_Persistent__*` prefix could be used to store information that e.g. C implementation stores inside structs which absolutely needs to survive ghostification.

Does it maybe help? (Sorry once again for maybe missing some your points - I've only briefly grasped your messages being short of time)

P.S.

> This doesn't happen in the C implementation because ghostify is only called from _p_deactivate and doesn't interact with the picklecache and new_ghost at all. The cPickleCache simply directly sets the object's state to GHOST.

To me this tells there are incompatibilities in beetween C- and Python implementations and the latter one better get fixed to behave compatible.

jamadden:
>  in pure-python case from slots that are cleared _p_* and _Persistent__* names are ommitted. I guess _Persistent__* prefix could be used to store information that e.g. C implementation stores inside structs which absolutely needs to survive ghostification.

> Does it maybe help? (Sorry once again for maybe missing some your points - I've only briefly grasped your messages being short of time)

Unfortunately, I don't think that helps. Most importantly, it still requires every class broken by this to change (in potentially complicated ways if they want to preserve compatibility). Also, `_Persistent__` names are, by convention, private names reserved to the implementation of the `Persistent` class; likewise, `_p_` names are (as I understand it) generally meant to be names used in the public implementation of the persistent machinery. Neither of those should be used casually by subclasses.

>> This doesn't happen in the C implementation because ghostify is only called from _p_deactivate and doesn't interact with the picklecache and new_ghost at all. The cPickleCache simply directly sets the object's state to GHOST.

> To me this tells there are incompatibilities in beetween C- and Python implementations and the latter one better get fixed to behave compatible.

That was only part of the story. That was the `new_ghost` part, and that part was possible to fix in the branch I referenced. But the rest of the story (the `_p_deactivate` part) is that it is possible to have completely private attributes in C (by simply putting them in the object struct) that are not visible in the `__dict__` or `__slot__` list; that's why `ghostify()` even now does not reach the `_wrapped` attribute of the C proxy (because it's not actually an attribute, it's a private member of the struct).

navytux:
I was talking about `_Persistent__` wrt making the logic of Persistent and cache interaction the same as in C version. About private members I now see. Python world used slots as a way to completely hide from persistence ghostification machinery and now since slots are ghostified to everything breaks.

I understand the rationale for not ghostifying slotted wars in pure-python case, but then it will create even more incompatibility in between C & Python implementations. Not releasing slotted memory in C case is also not a good way in my view.

So I can propose this: we can adjust Python implementation this way:

- slotted vars are released by default on ghostification
- if some class wants to omit some of its slotted variables from being release, it can do something like this:

```
class C(Persistent):
    __slots__ = ['attr1', 'attr2', ...]
    __keep__ = ['attr1']
```

the `__keep__` will tell persistency machinery not to release `attr1` on ghostification.

This way the fix for updating code to newer persistent release is straighforward (just add `__keep__ = __slots__`) and then those attributes that don't really need to be preserved can be excluded from `__keep__` on a case-by-case basis incrementally.

What do you think?

jamadden:
I am currently -1 on breaking things by default and being required to opt-in to the old behaviour. (But I'm feeling particularly curmudgeonly right now 😄) Such a break would require a coordinated release of persistent, BTrees, and zope.container (and that's only the projects I've looked at). As far as other projects go, it's particularly bad because the breakage might be silent: not every case will result in AttributeError or the like (e.g., an object that can read several versions of old states might have some `hasattr(self, 'foo')` calls sprinkled around).

If we're discussing opt-in/opt-out behaviour, then my counterproposal would be to have a class that *wants* its `__slots__` cleared to simply override `_p_deactivate`. That opt-in behaviour is possible today and has been for a long time. We could add an API or attribute to simplify that, even (without overriding `_p_deactivate`), but it would still be opt-in.

As I noted [over here](https://github.com/zopefoundation/BTrees/pull/47/files#r85588728), even clearing `__slots__` isn't a catch-all for objects that are implemented in C and store state in their struct, such as the zope.container proxy. To me, that suggests that the reliable, implicit rule that's been in use so far is "the `__dict__` goes, everything else is the object's responibility; override _p_deactivate to customize." I don't know how many projects are relying on that rule, but two out of a sample size of the four project's that I've run (BTrees, ZODB, ZEO, zope.container) is not a good ratio 😄

navytux:
@jamadden I see your point. Though to me it looks like having pure-python version along the canonical C-one is holding us to fix the long-stanging slots are not released issue. Said that I'm ok with any direction taken as long as C-version continues to release slotted attributes. Probably we can return to just doing major version bump as was originally proposed and that's all.

P.S.

And @jimfulton to me slots are not like `__setattr__/__getattr__` override - it is simply a way to save memory - when you have a class and put all used attributes in slots - there is no `__dict__` created and place for attributes is allocated directly alongside head object for an instance.

jamadden:
> Though to me it looks like having pure-python version along the canonical C-one is holding us to fix the long-stanging slots are not released issue.

Both implementations are of equal importance.

> Said that I'm ok with any direction taken as long as C-version continues to release slotted attributes. Probably we can return to just doing major version bump as was originally proposed and that's all.

I don't think I understand what you're suggesting here. It sounds like you're advocating to release two different implementations that behave differently, and which would still have the possibility to break things for users of the C extension...but not the Python version. That sounds like a nightmare.

> slots are not like __setattr__/__getattr__ override - it is simply a way to save memory - when you have a class and put all used attributes in slots - there is no __dict__ created and place for attributes is allocated directly alongside head object for an instance.

That's a good reason to try PyPy! It generally [uses less memory](https://morepypy.blogspot.com/2009/10/gc-improvements.html) than CPython due to its GC design, and it also auto-magically [makes normal `__dict__` instances as efficient as using `__slots__`](https://morepypy.blogspot.com/2010/11/efficiently-implementing-python-objects.html).

navytux:
@jamadden thanks for feedback. Sorry my last comment yesterday was more like thinking out in loud being too sleepy. I was trying to say: given the alternatives probably the best is to do a major version bump for persistent, as you were proposing in the first place and that's all. For sure we don't want python & C versions to become incompatble. Sorry for being very unclear.

About PyPy: probably it is not so visible but I've been with Python for a long time - since 2001. There are many good things in python, but over time I've realized several major blocking drawbacks are there: python being too slow, the GIL, and python being too dynamic (the last one makes doing refactoring or managing not small projects harder). Things are all inter-related here, e.g. dynamism induces slowness and GIL. Imho what could be a good direction to evolve is adopt ideas from Pyrex (now it is Cython) about static typing and this way remove unneccessary dynamism and improve code generation eventually paving the way for GIL to become unnecessary.  

PyPy tries to solve only speed issue (and only to some extent) but GIL and "too dynamic" stays there (I know about various attemts like transactional memory, but still practically GIL is still there). I can say many good words about PyPy, but since it also looses real C-API compatibiliy with CPython it really limits PyPy adoption in many projects(*). Since PyPy does not solve all Python problems, I personally switched into mode of using Python where we have to (older projects & compatibility requirements) and doing new things in Go, which to me can be seen as "better Python". And btw, @jimfulton you can have things like exceptions in go:

https://lab.nexedi.com/kirr/git-backup/blob/7535343c/error.go
https://lab.nexedi.com/kirr/git-backup/blob/7535343c/gitobjects.go#L103 (`raise`)
https://lab.nexedi.com/kirr/git-backup/blob/7535343c/git-backup.go#L979 (`except`)

though I also see good reasons to handle errors as values especially the more goroutiny a program is. In other words it is always a balance.

Kirill

(*) FYI at Nexedi we are also playing a bit with Pyston, since it can have C-API compatibility with CPython and already support NumPy/SciPy and Scikit-learn (and eventually hopefully Zope). Pyston is not there in terms of speedup yet (?) so for now the only real option for us is to use CPython.

jimfulton:
OK, I've looked over this and #48 (mostly this) and I thought and I though ... :)

I think this boils down to a difference between "intrinsic data" and state. Some notes:

https://github.com/zopefoundation/persistent/wiki/Notes_on_state_new_and_slots

I suggest the following compromise: only clear slots of an object's class doesn't override __new__.  This has the benefit of clearing slots when we know they only contain state.  IDK if this addresses the wendelin use case; hopefully it does, or can.

(I'm not sure if we still need #48, as I'm not sure what it fixes. )

jamadden:
I think that's a reasonable compromise. It fixes all the cases I know of that aren't outright bugs (zopefoundation/BTrees#47 is arguably such a bug, relying as it does on "intrinsic data" that really isn't). I wish I'd thought of it!

navytux:
@jimfulton thanks for suggesting this. What you propose is ok on my side (as we don't use `__new__`). We still need #48 because it fixes "releasing not initialized slot" case. I've implemented what you propse in	9137748 . I don't see how/whether it is possible to reopen merge request in github with more patches, so I've opened a new one: #50.

jimfulton:
Thanks!

jamadden:
@navytux Also thanks! It was a great discussion and I enjoyed it!

navytux:
@jimfulton, @jamadden - you are welcome.

navytux:
and thanks to you too.

PR #46, opened 2016-09-28
Make calls to unghost a cPersistent.Persistent object available in python

stephan-hof:
Hello,

I had a look in porting https://github.com/zopefoundation/Persistence to python3.
@hannosch added already support for the python version of persistent.
However we have seen problems if the C-Extensions of persistent is available.
In that case the python version of Persistence needs access to the C-code of persistent to unghostify objects (per_use, allow_deactivation, accessed)

My goal would be to implement  **getattribute** like this.

```
    def __getattribute__(self, key):

        if (not name.startswith('_p_') and name not in _SPECIAL_NAMES):
            persistent.PER_USE(self)
            v = ExtensionClass.Base_getattro(self, key)
            persistent.PER_ALLOW_DEACTIVATION(self)
            persistent.PER_ACCESSED(self)
        else:
            v = ExtensionClass.Base_getattro(self, key)
        return v
```

This pull request simply exposes some of the calls in cPersistence.h to python.

I know in a ideal world we would use the C-Extension of Persistence, so persistent.cPersistence.h can be used directly.
Unfortunately Persistence uses the C-Extension of ExtensionClass which is not ported to python3 (yet?).

If you don't like this pull request at all, I could add a mini C-Extension to Persistence which exposes the calls of persistent.cPersistence.h to python too.

### DIFF
--- a/persistent/__init__.py
+++ b/persistent/__init__.py
@@ -26,6 +26,9 @@ if not PURE_PYTHON:
         from persistent.cPersistence import CHANGED
         from persistent.cPersistence import STICKY
         from persistent.cPersistence import simple_new
+        from persistent.cPersistence import PER_USE
+        from persistent.cPersistence import PER_ALLOW_DEACTIVATION
+        from persistent.cPersistence import PER_ACCESSED
     except ImportError: #pragma NO COVER
         from persistent.persistence import Persistent
         from persistent.persistence import GHOST
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -1437,11 +1437,60 @@ simple_new(PyObject *self, PyObject *type_object)
     return PyType_GenericNew((PyTypeObject *)type_object, NULL, NULL);
 }
 
+static PyObject*
+py_per_use(PyObject* self, PyObject* ob) {
+    if (!PER_TypeCheck(ob)) {
+        return PyErr_Format(
+            PyExc_TypeError,
+            "object of type %s does not inherit from cPersistence.Persistent",
+            Py_TYPE(ob)->tp_name
+        );
+    }
+
+    if (PER_USE((cPersistentObject*) ob)) {
+        return NULL;
+    }
+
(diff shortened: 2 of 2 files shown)


stephan-hof:
Now that the C-Extension of ExtensionClass is ported to python3, Persistence can use the C API of persistent. So from my point of view this pull request is not needed anymore.

icemac:
@stephan-hof Could it still be helpful to support PyPy in future?

jamadden:
In the PyPy case, *none* of the C extensions would be used. If I understand this report correctly, the problem was about mixing C extensions of one project with Python versions of another, so the issue couldn't arise there.

Whether such a mixed scenario is something that should be supported is probably another discussion  (my thought: no thank you).

jimfulton:
OMG, I'm so out of touch with Zope, but WTF is Persistence still a thing?

Why doesn't Zope just use persistent?

OK, I guess this has something to do with "ExtensionClass semantics" and maybe Acquisition.  I think the Zope 4 effort needs to take a step back and consider dropping some of this legacy.

hannosch:
@jimfulton I tried pushing towards dropping more of the legacy code as part of the Zope 4 project. But the people who are left and willing to contribute to Zope are now those with large old applications incl. persistent code and lots of Acquisition magic. Basically Zope 3 and its approaches never happened to them.

That's of course of gross over-simplification, but the project goes where the active contributors want it to go. And these days that means the only thing that we can agree on is Python 3 compatibility and maintaining backwards compatibility with ZMI-driven development.

jimfulton:
I get and **endorse** a minimalist/incremental approach.

At some point though, the weight of the legacy code will increase the work of the few.

I would think that the getting rid of these would be a lot less effort than "Zope 3", but I'm probably forgetting things.  I wonder if it could be done with minor incompatibilities that people with legacy applications could deal with.  

But maybe It's not worth the bother. I'm not really in a position to judge the tradeoffs.

PR #47, opened 2016-10-27
Stop calling gc.collect in the Python incrgc.

jamadden:
@jimfulton and I have talked about it, and I'm (mostly :) convinced that
this shouldn't be an actual problem for any of the reasons described in
the previous comment.

Running the GC every time slows down one of Jim's benchmarks by up to 10x. It all depends on the frequency of the transactions relative to application work.

If I use PyPy 5.4.1 to run the ZODB master test suite against
this (well, with #44 rolled back) I don't get any unexpected
failures. (I haven't run the ZEO test suite yet.) Which honestly amazes
me because I'm sure I used to get test failures---I guess the PyPy GC
has changed...which means we may see some failures on Travis. I'll try
to set up an older PyPy to verify.

(ZODB tests won't run with current master until #44 is addressed somehow.)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,12 @@
 4.2.2 (2016-11-29)
 ------------------
 
+- Stop calling ``gc.collect`` every time ``PickleCache.incrgc`` is called (every
+  transaction boundary) in pure-Python mode (PyPy). This means that
+  the reported size of the cache may be wrong (until the next GC), but
+  it is much faster. This should not have any observable effects for
+  user code.
+
 - Drop use of ``ctypes`` for determining maximum integer size, to increase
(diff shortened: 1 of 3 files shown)


jamadden:
Appveyor tests have been failing since #31 was merged.

jamadden:
The Mac builds are failing to even start because of homebrew problems. I saw this on another zope project IIRC. I have no idea what causes it or what to do about it.

jamadden:
I ran the ZODB test suite with PyPy 2.5.0 and no issues. I couldn't run the ZEO test suite with that old of a PyPy but it ran fine with PyPy2 5.4.1.

jamadden:
Fixed by #54

jimfulton:
Why was this closed?

jimfulton:
Oh, nm

PR #48, opened 2016-10-28
py/deactivate vs slots: A slot variable could not be initialized at all

navytux:
fe2219f4 (On deactivate release in-slots objects too) started to release
objects from slotted variables but was not careful enough while doing so: we
have to be careful while deleting as for unset variables it will raise
AttributeError:

```
class C(object):
    __slots__ = ['aaa', 'unset']

    def __init__(self, aaa):
        self.aaa = 1

c = C(111)
del c.aaa
del c.unset

  File "y.py", line 9, in <module>
    del c.unset
AttributeError: unset
```

Caught by preparing fix for https://github.com/zopefoundation/persistent/pull/44#issuecomment-256768600

C version already have this protection in the original fe2219f4 patch:

```
+            if (PyObject_GenericSetAttr((PyObject *)self, name, NULL) < 0)
+                /* delattr of non-set slot will raise AttributeError - we
+                 * simply ignore. */
+                PyErr_Clear();
```

### DIFF
--- a/persistent/persistence.py
+++ b/persistent/persistence.py
@@ -428,7 +428,12 @@ class Persistent(object):
         #   class does not override __new__ )
         if type_.__new__ is Persistent.__new__:
             for slotname in Persistent._slotnames(self, _v_exclude=False):
-                getattr(type_, slotname).__delete__(self)
+                try:
+                    getattr(type_, slotname).__delete__(self)
+                except AttributeError:
+                    # AttributeError means slot variable was not initialized at all -
+                    # - we can simply skip its deletion.
+                    pass
 
         # Implementation detail: deactivating/invalidating
         # updates the size of the cache (if we have one)
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -1321,7 +1321,7 @@ class _Persistent_Base(object):
 
     def test__p_invalidate_from_changed_w_slots(self):
         class Derived(self._getTargetClass()):
-            __slots__ = ('myattr1', 'myattr2')
+            __slots__ = ('myattr1', 'myattr2', 'unset')
             def __init__(self):
                 self.myattr1 = 'value1'
                 self.myattr2 = 'value2'


navytux:
about CI being red:

- for travis it is only several pre-install problems, like "Error: python-2.7.12_1 already installed", e.g.  https://travis-ci.org/zopefoundation/persistent/jobs/171338522 or "/Users/travis/build.sh: line 109: shell_session_update: command not found", e.g. https://travis-ci.org/zopefoundation/persistent/jobs/171338524
- for appveyour it is only C & Py hashes being different in 64bit mode, but it was already so before, e.g. here: https://ci.appveyor.com/project/mgedmin/persistent/build/1.0.13/job/9f374y4jfxic26c4

I mean there are no new failures.

navytux:
Ping. This is a fix for pure-python part of fe2219f4 and imho should be thus applied too.

/cc @jimfulton, @tseaver

jamadden:
I believe the entire change should be rolled back as it breaks existing Python code with *no BWC way to fix it* for certain use cases (of note, zope.container). Further, this behavior is easily opted into for those objects that choose too (and note that some C objects do not choose to, even with this change, but Python has no choice). See the comments beginning at https://github.com/zopefoundation/persistent/pull/44#issuecomment-256920566 for an in depth exploration.

navytux:
Whatever way it is either #44 and this PR should be applied or both not applied. Btw I thought we settled on having just major version change bump and that's all.

jamadden:
>  Btw I thought we settled on having just major version change bump and that's all.

That was the last thing you commented, based on a comment I made way at the beginning of the discussion, before I understood all of the implications. Once I understood that it's simply not possible to make zope.container work with this change without invalidating all existing databases that use it, I stopped advocating that. (I'd still be happy to be proven wrong about that.)

The backwards compatibility bar for persistent is very high due to existing data.

jimfulton:
Please see the comment I just made on #44 .

navytux:
(rebased on top of #50)

navytux:
(once again rebased on top of adjusted #50)

navytux:
(once again rebased on top of adjusted #50)

jimfulton:
For some reason, this isn't showing diffs relative to current master. :( 

Anyone know wtf?

jimfulton:
Much better. Thanks.

jimfulton:
This looks OK to me.  @jamadden I assume you're cool with this now.

I'd like to make a release today if possible.

jamadden:
LGTM

navytux:
@jimfulton, @jmadden - thanks for applying the patch and for feedback.

PR #50, opened 2016-11-16
On deactivate release in-slots objects only for classes that don't ov…

navytux:
…erride __new__

Commit fe2219f4 taught Persistent to release in-slots objects on
deactivation. That however broke pure-python implementation of many
things because they were using __slots__ as a place which survive
deactivation.

As per discussion in

    https://github.com/zopefoundation/persistent/pull/44

and resolution:

    https://github.com/zopefoundation/persistent/pull/44#issuecomment-261019084

let's try to preserve backward compatibility by not releasing slots for
classes that override __new__

/proposed-by @jimfulton

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -190,47 +190,52 @@ ghostify(cPersistentObject *self)
         *dictptr = NULL;
     }
 
-    /* clear all slots besides _p_* */
-    slotnames = pickle_slotnames(Py_TYPE(self));
-    if (slotnames && slotnames != Py_None)
+    /* clear all slots besides _p_*
+     * ( for backward-compatibility reason we do this only if class does not
+     *   override __new__ ) */
+    if (Py_TYPE(self)->tp_new == Pertype.tp_new)
     {
-        int i;
-
-        for (i = 0; i < PyList_GET_SIZE(slotnames); i++)
+        slotnames = pickle_slotnames(Py_TYPE(self));
+        if (slotnames && slotnames != Py_None)
         {
-            PyObject *name;
-            char *cname;
-            int is_special;
+            int i;
 
-            name = PyList_GET_ITEM(slotnames, i);
-#ifdef PY3K
-            if (PyUnicode_Check(name))
+            for (i = 0; i < PyList_GET_SIZE(slotnames); i++)
             {
-                PyObject *converted = convert_name(name);
-                cname = PyBytes_AS_STRING(converted);
-#else
-            if (PyBytes_Check(name))
-            {
-                cname = PyBytes_AS_STRING(name);
-#endif
-                is_special = !strncmp(cname, "_p_", 3);
+                PyObject *name;
+                char *cname;
+                int is_special;
+
+                name = PyList_GET_ITEM(slotnames, i);
 #ifdef PY3K
-                Py_DECREF(converted);
-#endif
-                if (is_special) /* skip persistent */
+                if (PyUnicode_Check(name))
                 {
-                    continue;
+                    PyObject *converted = convert_name(name);
+                    cname = PyBytes_AS_STRING(converted);
+#else
+                if (PyBytes_Check(name))
+                {
+                    cname = PyBytes_AS_STRING(name);
+#endif
+                    is_special = !strncmp(cname, "_p_", 3);
+#ifdef PY3K
+                    Py_DECREF(converted);
+#endif
+                    if (is_special) /* skip persistent */
+                    {
+                        continue;
+                    }
                 }
-            }
 
-            /* NOTE: this skips our delattr hook */
-            if (PyObject_GenericSetAttr((PyObject *)self, name, NULL) < 0)
-                /* delattr of non-set slot will raise AttributeError - we
-                 * simply ignore. */
-                PyErr_Clear();
+                /* NOTE: this skips our delattr hook */
+                if (PyObject_GenericSetAttr((PyObject *)self, name, NULL) < 0)
+                    /* delattr of non-set slot will raise AttributeError - we
+                     * simply ignore. */
+                    PyErr_Clear();
+            }
         }
+        Py_XDECREF(slotnames);
     }
-    Py_XDECREF(slotnames);
 
     /* We remove the reference to the just ghosted object that the ring
     * holds.  Note that the dictionary of oids->objects has an uncounted
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -1340,6 +1340,36 @@ class _Persistent_Base(object):
         self.assertEqual(list(jar._loaded), [])
         self.assertEqual(list(jar._registered), [])
 
+    def test__p_invalidate_from_changed_w_slots_compat(self):
+        # check that (for backward-compatibility reason) slots are not released
+        # for classes where __new__ is overwritten. Attributes in __dict__
+        # should be always released.
+        class Derived(self._getTargetClass()):
+            __slots__ = ('myattr1', 'myattr2', '__dict__')
+            def __new__(cls):
+                obj = cls.__base__.__new__(cls)
+                obj.myattr1 = 'value1'
+                obj.myattr2 = 'value2'
+                obj.foo = 'foo1' # .foo & .bar are in __dict__
+                obj.bar = 'bar2'
+                return obj
+        inst, jar, OID = self._makeOneWithJar(Derived)
+        inst._p_activate()
+        inst._p_changed = True
+        jar._loaded = []
+        jar._registered = []
+        self.assertEqual(Derived.myattr1.__get__(inst), 'value1')
+        self.assertEqual(Derived.myattr2.__get__(inst), 'value2')
+        self.assertEqual(inst.__dict__, {'foo': 'foo1', 'bar': 'bar2'})
+        inst._p_invalidate()
+        self.assertEqual(inst._p_status, 'ghost')
+        self.assertEqual(list(jar._loaded), [])
+        self.assertEqual(Derived.myattr1.__get__(inst), 'value1')
+        self.assertEqual(Derived.myattr2.__get__(inst), 'value2')
+        self.assertEqual(inst.__dict__, {})
+        self.assertEqual(list(jar._loaded), [])
+        self.assertEqual(list(jar._registered), [])
+
     def test__p_invalidate_from_sticky(self):
         inst, jar, OID = self._makeOneWithJar()
         inst._p_activate() # XXX
--- a/persistent/persistence.py
+++ b/persistent/persistence.py
@@ -424,8 +424,11 @@ class Persistent(object):
         if idict is not None:
             idict.clear()
         type_ = type(self)
-        for slotname in Persistent._slotnames(self, _v_exclude=False):
-            getattr(type_, slotname).__delete__(self)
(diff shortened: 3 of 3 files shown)


navytux:
(CI being red is the same x86-64 hashes differences that was before this patch)

persistent/tests/test_persistence.py
```diff
@@ -1340,6 +1340,31 @@ def __init__(self):
         self.assertEqual(list(jar._loaded), [])
         self.assertEqual(list(jar._registered), [])
 
+    def test__p_invalidate_from_changed_w_slots_compat(self):
+        # check that (for backward-compatibility reason) slots are not released
+        # for classes where __new__ is overwritten.
+        class Derived(self._getTargetClass()):
+            __slots__ = ('myattr1', 'myattr2')
```
jamadden: It might be a good idea to add `__dict__` and set an attribute in it that's *not* in `__slots__` and then confirm that `__dict__` still gets cleared even when the slots aren't. (That's the circumstance that the contained proxy is in.)

navytux: good idea.

navytux:
Patch updated to verify that __dict__ is always cleared irreagardless of whether `__new__` is overwritten or not. Interdiff:

```diff
diff --git a/persistent/tests/test_persistence.py b/persistent/tests/test_persistence.py
index 070381c..0922b64 100644
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -1342,13 +1342,16 @@ def __init__(self):

     def test__p_invalidate_from_changed_w_slots_compat(self):
         # check that (for backward-compatibility reason) slots are not released
-        # for classes where __new__ is overwritten.
+        # for classes where __new__ is overwritten. Attributes in __dict__
+        # should be always released.
         class Derived(self._getTargetClass()):
-            __slots__ = ('myattr1', 'myattr2')
+            __slots__ = ('myattr1', 'myattr2', '__dict__')
             def __new__(cls):
                 obj = cls.__base__.__new__(cls)
                 obj.myattr1 = 'value1'
                 obj.myattr2 = 'value2'
+                obj.foo = 'foo1' # .foo & .bar are in __dict__
+                obj.bar = 'bar2'
                 return obj
         inst, jar, OID = self._makeOneWithJar(Derived)
         inst._p_activate()
@@ -1357,11 +1360,13 @@ def __new__(cls):
         jar._registered = []
         self.assertEqual(Derived.myattr1.__get__(inst), 'value1')
         self.assertEqual(Derived.myattr2.__get__(inst), 'value2')
+        self.assertEqual(inst.__dict__, {'foo': 'foo1', 'bar': 'bar2'})
         inst._p_invalidate()
         self.assertEqual(inst._p_status, 'ghost')
         self.assertEqual(list(jar._loaded), [])
         self.assertEqual(Derived.myattr1.__get__(inst), 'value1')
         self.assertEqual(Derived.myattr2.__get__(inst), 'value2')
+        self.assertEqual(inst.__dict__, {})
         self.assertEqual(list(jar._loaded), [])
         self.assertEqual(list(jar._registered), [])
```

navytux:
@jimfulton, @jamadden am I right we are on right track here? For C part most of the patch is only indent change so with `git show -w` it looks like:

```diff
diff --git a/persistent/cPersistence.c b/persistent/cPersistence.c
index 5a5bfc2..3fa2566 100644
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -34,6 +34,7 @@ static PyObject *py___getattr__, *py___setattr__, *py___delattr__;
 static PyObject *py___slotnames__, *copy_reg_slotnames, *__newobj__;
 static PyObject *py___getnewargs__, *py___getstate__;
 static PyObject *py_unsaved, *py_ghost, *py_saved, *py_changed, *py_sticky;
+static PyObject *py___new__, *Persistent__new__;
 
 
 static int
@@ -59,6 +60,7 @@ init_strings(void)
   INIT_STRING(saved);
   INIT_STRING(changed);
   INIT_STRING(sticky);
+  INIT_STRING(__new__);
 #undef INIT_STRING
   return 0;
 }
@@ -151,7 +153,7 @@ accessed(cPersistentObject *self)
 static void
 ghostify(cPersistentObject *self)
 {
-    PyObject **dictptr, *slotnames;
+    PyObject **dictptr, *slotnames, *__new__;
 
     /* are we already a ghost? */
     if (self->state == cPersistent_GHOST_STATE)
@@ -190,7 +192,13 @@ ghostify(cPersistentObject *self)
         *dictptr = NULL;
     }
 
-    /* clear all slots besides _p_* */
+    /* clear all slots besides _p_*
+     * ( for backward-compatibility reason we do this only if class does not
+     *   override __new__ ) */
+    __new__ = PyObject_GetAttr((PyObject *)Py_TYPE(self), py___new__);
+    Py_XDECREF(__new__);
+    if (__new__ == Persistent__new__)
+    {
         slotnames = pickle_slotnames(Py_TYPE(self));
         if (slotnames && slotnames != Py_None)
         {
@@ -231,6 +239,7 @@ ghostify(cPersistentObject *self)
             }
         }
         Py_XDECREF(slotnames);
+    }
 
     /* We remove the reference to the just ghosted object that the ring
     * holds.  Note that the dictionary of oids->objects has an uncounted
@@ -1501,6 +1510,9 @@ module_init(void)
     Pertype.tp_new = PyType_GenericNew;
     if (PyType_Ready(&Pertype) < 0)
         return NULL;
+    Persistent__new__ = PyDict_GetItem(Pertype.tp_dict, py___new__);
+    if (!Persistent__new__)
+        return NULL;
     if (PyModule_AddObject(module, "Persistent", (PyObject *)&Pertype) < 0)
         return NULL;
```

I mean is it ok or maybe there are some reasons to not apply it?

Thanks beforehand,
Kirill

jamadden:
I'm not a C-API expert so I'll leave that to others (although it *looks* ok to me). The Python implementation appears to be correct (but I haven't personally tested it with zope.container yet).

navytux:
@jamadden thanks for feedback. So let's wait for your zope.container testing (since the patch is solely purposed to recover backward-compatibility for it & similar codes) and then, if it works there, for C-part review.

Hope it will be ok.

jamadden:
With #50 and #48 applied to current master of persistent, zope.container tests pass in a pure-Python environment.

navytux:
@jamadden thanks for feedback.

@jimfulton is the C-part ok? should we merge patches now?

persistent/cPersistence.c
```diff
@@ -190,47 +192,54 @@ ghostify(cPersistentObject *self)
         *dictptr = NULL;
     }
 
-    /* clear all slots besides _p_* */
-    slotnames = pickle_slotnames(Py_TYPE(self));
-    if (slotnames && slotnames != Py_None)
+    /* clear all slots besides _p_*
+     * ( for backward-compatibility reason we do this only if class does not
+     *   override __new__ ) */
+    __new__ = PyObject_GetAttr((PyObject *)Py_TYPE(self), py___new__);
```
jimfulton: You don't need and shouldn't use (because overhead) getattr here.  You can just compare the tp_new struct members:

    #ifdef PY3K
        if (self->ob_base.ob_type->tp_new == Pertype.tp_new)
    #else
        if (self->ob_type->tp_new == Pertype.tp_new)
    #endif

This will simplify this change quite a bit, as you won't need the __new__string or to cache the Python version of persistent.__new__.

navytux: @jimfulton thanks for your feedback. You are right. I originally started the patch the way you suggest but it was not working here the way it should - probably some thinko on my side that `__new__` for pythonic objects stored only in type's `tp_dict` and not directly in `tp_new`. I clearly remember I was getting tests failures but now no more; strange... Thanks for catching this.

Updating the patch with the following interdiff:

```diff
diff --git a/persistent/cPersistence.c b/persistent/cPersistence.c
index 3fa2566..67cdbd2 100644
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -34,7 +34,6 @@ static PyObject *py___getattr__, *py___setattr__, *py___delattr__;
 static PyObject *py___slotnames__, *copy_reg_slotnames, *__newobj__;
 static PyObject *py___getnewargs__, *py___getstate__;
 static PyObject *py_unsaved, *py_ghost, *py_saved, *py_changed, *py_sticky;
-static PyObject *py___new__, *Persistent__new__;
 
 
 static int
@@ -60,7 +59,6 @@ init_strings(void)
   INIT_STRING(saved);
   INIT_STRING(changed);
   INIT_STRING(sticky);
-  INIT_STRING(__new__);
 #undef INIT_STRING
   return 0;
 }
@@ -153,7 +151,7 @@ accessed(cPersistentObject *self)
 static void
 ghostify(cPersistentObject *self)
 {
-    PyObject **dictptr, *slotnames, *__new__;
+    PyObject **dictptr, *slotnames;
 
     /* are we already a ghost? */
     if (self->state == cPersistent_GHOST_STATE)
@@ -195,9 +193,7 @@ ghostify(cPersistentObject *self)
     /* clear all slots besides _p_*
      * ( for backward-compatibility reason we do this only if class does not
      *   override __new__ ) */
-    __new__ = PyObject_GetAttr((PyObject *)Py_TYPE(self), py___new__);
-    Py_XDECREF(__new__);
-    if (__new__ == Persistent__new__)
+    if (Py_TYPE(self)->tp_new == Pertype.tp_new)
     {
         slotnames = pickle_slotnames(Py_TYPE(self));
         if (slotnames && slotnames != Py_None)
@@ -1510,9 +1506,6 @@ module_init(void)
     Pertype.tp_new = PyType_GenericNew;
     if (PyType_Ready(&Pertype) < 0)
         return NULL;
-    Persistent__new__ = PyDict_GetItem(Pertype.tp_dict, py___new__);
-    if (!Persistent__new__)
-        return NULL;
     if (PyModule_AddObject(module, "Persistent", (PyObject *)&Pertype) < 0)
         return NULL;
```

(with `Py_TYPE` we don't have to ifdef)

Hope the patch is better now. With `git show -w` for C part it is now only:

```diff
diff --git a/persistent/cPersistence.c b/persistent/cPersistence.c
index 5a5bfc2..67cdbd2 100644
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -190,7 +190,11 @@ ghostify(cPersistentObject *self)
         *dictptr = NULL;
     }
 
-    /* clear all slots besides _p_* */
+    /* clear all slots besides _p_*
+     * ( for backward-compatibility reason we do this only if class does not
+     *   override __new__ ) */
+    if (Py_TYPE(self)->tp_new == Pertype.tp_new)
+    {
         slotnames = pickle_slotnames(Py_TYPE(self));
         if (slotnames && slotnames != Py_None)
         {
@@ -231,6 +235,7 @@ ghostify(cPersistentObject *self)
             }
         }
         Py_XDECREF(slotnames);
+    }
 
     /* We remove the reference to the just ghosted object that the ring
     * holds.  Note that the dictionary of oids->objects has an uncounted
```

Thanks again,
Kirill

navytux:
(CI is red, once again, only on x86-64 AppVeyor because of hash differences that were there before my patch)

jimfulton:
It's interesting that 2 of the 3 mac builds are green now.  wtf.  The last is waiting.

I wish we had MS devs. I'm trying to set up an EC2 MS instance...

navytux:
@jimfulton thanks for merging the patch.

PR #53, opened 2016-12-16
Expose PersistentList/Mapping docs. Fixes #52.

### LINKED ISSUE #52: Docs: Document PersistentMapping and PersistentList
These aren't referenced from the local docs anywhere. (They're pulled into the ZODB docs on zodb.readthedocs.io for some reason.) Some narrative on why you'd want to use them might be appropriate here.

jamadden:


docs/api.rst
```diff
@@ -5,6 +5,7 @@
    :maxdepth: 2
 
    api/interfaces
+   api/collections
```
tseaver: This file seems to be missing from the PR.

docs/using.rst
```diff
@@ -412,33 +417,34 @@ Overriding the attribute protocol
 Subclasses which override the attribute-management methods provided by
 :class:`persistent.Persistent`, but must obey some constraints:
 
-:meth:`__getattribute__``
+
+``__getattribute__``
```
tseaver: Why are we dropping the `:meth:` role?

persistent/list.py
```diff
@@ -21,6 +21,12 @@
 from persistent._compat import PYTHON2
 
 class PersistentList(UserList, persistent.Persistent):
+    """
```
tseaver: Summary line follows the triple quote immediately (PEP 257).

docs/using.rst
tseaver: I don't know from that tool.  It looks as though `:method:` is the [currenly-in-vogue spelling](http://www.sphinx-doc.org/en/stable/domains.html#directive-py:method).

jamadden: It just runs sphinx with the warnings cranked up is all.

But that change is really unrelated, I'll revert.

jamadden: I suspect probably what it was complaining about was that the :meth: appeared outside of a class body, and [without a leading class name](http://www.sphinx-doc.org/en/stable/domains.html#directive-py:class). So it didn't know what it was trying to describe.

docs/conf.py
```diff
@@ -50,16 +50,16 @@
...
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The short X.Y version.
-version = '3.10'
+version = '4.2'
 # The full version, including alpha/beta/rc tags.
-release = '3.10b1'
+release = '4.2.2'
```
icemac: This is not DRY. It was forgotten to update for years. The only way to make it DRY seems to be using this `version.txt` approach.

PR #54, opened 2016-12-19
new_ghost shouldn't clear newargs (PyPy consistent with C)

### LINKED ISSUE #49: PyPy/Pure-Python: `PickleCache.new_ghost` destroys any information passed to `__new__`
Spurred from the work in #44, with a lengthy description in https://github.com/zopefoundation/persistent/pull/44#issuecomment-256962199

Both dicts and slots are affected and need a test case.

jamadden:
Fixes #49. Tested with ZODB, ZEO, zope.container.

(Includes #47 , rebased on master and tested with PyPy on ZEO and ZODB, because when I created this branch I expected that PR to be merged already by the time this was ready.)

### DIFF
--- a/persistent/persistence.py
+++ b/persistent/persistence.py
@@ -422,8 +422,11 @@ class Persistent(object):
             _OSA(self, '_Persistent__flags', None)
 
         if clear:
-            idict = _OGA(self, '__dict__', None)
-            if idict is not None:
+            try:
+                idict = _OGA(self, '__dict__')
+            except AttributeError:
+                pass
+            else:
                 idict.clear()
             type_ = type(self)
             # for backward-compatibility reason we release __slots__ only if
--- a/persistent/persistence.py
+++ b/persistent/persistence.py
@@ -413,27 +413,29 @@ class Persistent(object):
         _OSA(self, '_Persistent__flags', 0)
         self._p_deactivate()
 
-    def _p_invalidate_deactivate_helper(self):
+    def _p_invalidate_deactivate_helper(self, clear=True):
         jar = _OGA(self, '_Persistent__jar')
         if jar is None:
             return
 
         if _OGA(self, '_Persistent__flags') is not None:
             _OSA(self, '_Persistent__flags', None)
-        idict = getattr(self, '__dict__', None)
-        if idict is not None:
-            idict.clear()
-        type_ = type(self)
-        # ( for backward-compatibility reason we release __slots__ only if
-        #   class does not override __new__ )
-        if type_.__new__ is Persistent.__new__:
-            for slotname in Persistent._slotnames(self, _v_exclude=False):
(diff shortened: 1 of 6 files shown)


persistent/tests/test_persistence.py
```diff
@@ -1627,6 +1675,10 @@ def update_object_size_estimation(self, oid, new_size):
 
         return _Cache(jar)
 
+    def _makeRealCache(self, jar):
+        from persistent.picklecache import PickleCache
```
jamadden: CPersistentTests already used cPickleCache in _makeCache, not a mock like the PyPersistentTests did.

persistent/persistence.py
```diff
@@ -413,27 +413,29 @@ def _p_invalidate(self):
...
+            # for backward-compatibility reason we release __slots__ only if
+            # class does not override __new__
+            if type_.__new__ is Persistent.__new__:
+                for slotname in Persistent._slotnames(self, _v_exclude=False):
+                    try:
+                        getattr(type_, slotname).__delete__(self)
+                    except AttributeError:
+                        # AttributeError means slot variable was not initialized at all -
+                        # - we can simply skip its deletion.
+                        pass
```
tseaver: ISTM that we need a protocol by which a class derived from `Persistent` can signal signal explicitly which attrs / slots should be preserved:  all the others *should* be cleared by `new_ghost` (a ghost isn't supposed to have any persistent state set).

jamadden: This was sort-of talked over in #44. I think the consensus was that some sort of protocol would be nice, BWC makes it difficult to do.

The change here just brings parity between the Python implementation and what the C implementation is already doing.

jamadden:
@tseaver Any further comments?

tseaver:
I'm still puzzled that the C version of `new_ghost` doesn't clear object state:  the whole `__getattr__`-triggers-load scheme kind of depends on it.

jamadden:
The only state an object would have when `new_ghost` is called by ZODB is what comes from `__getnewargs__`. Most of the C implemented objects store that information in private `struct` fields, not in slots or the dict (what Jim called "intrinsic data"), so it wouldn't have been cleared *anyway*. 

The situation for Python-implemented objects is more complex, but with the C cache implementation, nothing passed to `__new__` would have been cleared at `new_ghost` time. The Python cache implementation was behaving differently, breaking objects that were relying on that. 

I don't know any *actual* cases where this implementation difference has mattered, but the increasing complexity of how `__slots__` are handled---not ever being cleared if `__new__` is overridden---suggests that we at least need the test cases for this behaviour to make sure that it is consistent between the implementations and doesn't *accidentally* regress.

tseaver:
LGTM

jamadden:
Thanks for the review!

PR #55, opened 2016-12-19
Fix Python timestamp hash on 64-bit windows

### LINKED ISSUE #51: Appveyor failures for persistent 4.2.2 (test failures on 64-bit platforms)
AFAICT the latest persistent release has no Windows wheels on PyPI because https://ci.appveyor.com/project/mgedmin/persistent/build/1.0.31 is all red.

All 64-bit tests fail with

```
FAIL: test_hash_equal (persistent.tests.test_timestamp.PyAndCComparisonTests)
  File "c:\projects\persistent\persistent\tests\test_timestamp.py", line 266, in test_hash_equal
    self.assertEqual(hash(c), hash(py))
AssertionError: -1419374591 != 1979033151
FAIL: test_hash_equal_constants (persistent.tests.test_timestamp.PyAndCComparisonTests)
  File "c:\projects\persistent\persistent\tests\test_timestamp.py", line 334, in test_hash_equal_constants
    self.assertEqual(c.__hash__(), 1000006000001)
AssertionError: -721379967 != 1000006000001L
```
(except on Python 3 the 1st error is `AssertionError: -1419374591 != -3850693964765720575`).

I can't figure out why the 32-bit tests failed (an error from `twine upload`?).

jamadden:
Fixes #51

Part of the problem was in the tests. The detection of when the hash
gets truncated was broken on win32.

Part of the problem is somewhere in the _wraparound function, I think,
because even fixing the tests to use the right values we still get
mismatches. Rather than dig into that, I just go back to using ctypes
when it's available.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@
 ------------------
 
 - Drop use of ``ctypes`` for determining maximum integer size, to increase
-  pure-Python compatibility.
+  pure-Python compatibility. See https://github.com/zopefoundation/persistent/pull/31
 
 - Ensure that ``__slots__`` attributes are cleared when a persistent
   object is ghostified.  (This excluses classes that override
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 ``persistent`` Changelog
 ========================
 
+4.2.3 (unreleased)
+------------------
+
+- Fix the hashcode of Python ``TimeStamp`` objects on 64-bit Python on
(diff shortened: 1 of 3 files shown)


jamadden:
Nevermind, I found the problem in _wraparound and was able to avoid ctypes.

jamadden:
Argh! Windows went green, but PyPy3 broke. Going back to ctypes.

jamadden:
Yay, first green build in awhile!

PR #56, opened 2016-12-22
Test w/ PURE_PYTHON set under Python3.5.

tseaver:


jamadden:
Do we want any of the PURE_PYTHON variants to be tested on Travis? It doesn't use tox.ini.

PR #57, opened 2017-01-13
Support Python 3.6

mgedmin:


.travis.yml
```diff
@@ -11,6 +11,8 @@ matrix:
         - os: linux
           python: 3.5
         - os: linux
+          python: 3.6
+        - os: linux
```
jamadden: Any idea if 3.6 is available for the OS X builds? Since they upload wheels it would probably be good to add it if possible.

mgedmin: I believe so: https://travis-ci.org/zopefoundation/zope.index/builds/176093903 failed on OSX because "homebrew 3" is 3.6 already.

mgedmin: (I would prefer that to be in a separate PR, since I've zero practical experience with macs or homebrew)

jamadden: I run OS X and, other than a bunch of sign-comparison warnings that aren't there in 3.5, 3.6 tests clean. I have no experience with homebrew, though.

```
persistent/cPickleCache.c:526:43: warning: comparison of integers of different signs: 'Py_ssize_t' (aka 'long') and 'unsigned long' [-Wsign-compare]
                (v->ob_type->tp_basicsize >= sizeof(cPersistentObject))
                 ~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~
persistent/cPickleCache.c:714:39: warning: comparison of integers of different signs: 'Py_ssize_t' (aka 'long') and 'unsigned long' [-Wsign-compare]
    else if (v->ob_type->tp_basicsize < sizeof(cPersistentObject))
             ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~
persistent/cPickleCache.c:1036:39: warning: comparison of integers of different signs: 'Py_ssize_t' (aka 'long') and 'unsigned long' [-Wsign-compare]
    else if (v->ob_type->tp_basicsize < sizeof(cPersistentObject))
             ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~
```

mgedmin: For the record, these warnings show up on Linux as well (`tox -e py36`).

jamadden: This *appears* to be a change in the way the default CFLAGS from `_sysconfigdata` get used. I've verified that the typedefs for `tp_basicsize` haven't changed from 3.5 to 3.6 (at least on Mac). Both versions of `_sysconfigdata` (for me) specify `-Wsign-compare` which triggers the warning; that's only used, though, if I don't have CFLAGS set in my own environment (so I can trigger the warning on both 3.5 and 3.6). But when tox does the builds, I only get the warning for 3.6...so something has changed in the way the CFLAGS are set/used/cleared/something.

tl;dr: Not a new problem.

mgedmin:
:+1:

PR #58, opened 2017-01-13
Python 3.6 Mac

jamadden:
Added at the top to see if it would change the build order. Turns out it doesn't, they're grouped by OS and the linux builds go first (and faster). (I would move it down with the others, but then there'd have to be another build and it's taking more than an hour now for the mac builds to start.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: python
 sudo: false
 matrix:
     include:
-        - os: osx
-          language: generic
-          env: TERRYFY_PYTHON='homebrew 3.6'
         - os: linux
           python: 2.7
         - os: linux
@@ -27,7 +24,13 @@ matrix:
           env: TERRYFY_PYTHON='macpython 3.4'
         - os: osx
           language: generic
-          env: TERRYFY_PYTHON='homebrew 3.5'
+          env: TERRYFY_PYTHON='macpython 3.5'
+        # Homebrew only accepts the major version number and
+        # automatically rolls out new versions. At this writing, that
+        # is 3.6.0.
+        - os: osx
+          language: generic
+          env: TERRYFY_PYTHON='homebrew 3'
 before_install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then git clone https://github.com/MacPython/terryfy; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then source terryfy/travis_tools.sh; fi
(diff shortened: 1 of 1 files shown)


jamadden:
Sigh. Something's wrong. Not with this PR, with the OS X builds in general, I think:

I added `python --version` for assurances of what `python` does before `python setup.py test`, and it turns out that the `homebrew 3.5` build is [actually running Python 3.6](https://travis-ci.org/zopefoundation/persistent/jobs/191779292#L698). 

So now we have two 3.6 builds and no 3.5 build.

jamadden:
The [comments for get_python_environment](https://github.com/MacPython/terryfy/blob/master/travis_tools.sh#L237) imply that homebrew only pays attention to the major version number. So we may not be able to use homebrew to get 3.5 anymore.

I tried using macpython instead for 3.6, which is supposed to download the installer from python.org, but [it blew up spectacularly and weirdly](https://travis-ci.org/zopefoundation/persistent/jobs/191749081).  Maybe I'll try that for 3.5...

jamadden:
Ok, so 'macpython 3.5' and 'homebrew 3' get us Python 3.5 and 3.6, respectively, so this PR has things back in a good state. 

I *suspect*, but did not test, that 'macpython 3.6.0' would also get us 3.6 ('macpython 3.6' currently fails as mentioned above). That might be better than floating with whatever 'homebrew 3' decides to be.

mgedmin:
How many other zope packaged are no longer building python 3.5 wheels for mas os x due to this homebrew thing?

jamadden:
> How many other zope packaged are no longer building python 3.5 wheels for mas os x due to this homebrew thing?

BTrees, ExtensionClass and Acquisition are the only packages I can think of that include C extensions. Of those, only BTrees is building wheels on OS X, and it [appears to be impacted](https://pypi.python.org/pypi/BTrees). 

Which points out another problem: using a 'homebrew' version gets you the *wrong* wheel tag: you get the restrictive '10_11_x86_64' tag (64-bit only 10.11 or greater) instead of the very general '10_6_intel' (32 and 64-bits, 10.6 and above). So pretty much all wheels should be using 'macpython', not 'homebrew'.

Given how congested the Mac build farm is, hopefully we're not doing builds on it unless necessary (i.e., to upload C extension wheels)

mgedmin:
For the record

```
mg@platonas: ~/src/zopefoundation $ grep -l homebrew */.travis.yml
BTrees/.travis.yml
persistent/.travis.yml
zope.index/.travis.yml
zope.interface/.travis.yml
```

PR #59, opened 2017-01-14
Use macpython for all Mac builds.

jamadden:
Ref #58. This should get us the right wheel tags and hopefully the non-floating 3.6 build (until 'multibuild' from terryfy is updated we have to use a specific 3.6.0 version).

Fingers crossed.

EDIT: I forgot zope.interface from the list of things with C extensions.
EDIT2: Oh, and I guess zope.index has an *optional* extension.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,21 +16,21 @@ matrix:
           python: pypy
         - os: linux
           python: pypy3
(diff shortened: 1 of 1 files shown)


jamadden:
Other than the unpleasantness of needing to specify '3.6.0' (for now), this seems to do what we want.

PR #61, opened 2017-02-22
Fix ``__setstate__`` interning when ``state`` parameter is not a built-in dict

### LINKED ISSUE #60: Persistent.__setstate__ C SILENTLY fails on non-builtin dict
Key interning introduced in https://github.com/zopefoundation/persistent/commit/7bd85c189ae17be514f4989bfb8f79f4d8013943 makes ```__setstate__``` SILENTLY fail when a  non-builtin dict is passed as state.

See https://github.com/agroszer/persistent/commit/1c1cebaf79914420a9de61e55fb72a75981965b2 how to provoke that.

- is there any constraint that ```state``` must be a builtin dict ?
- the C and python implementation differs, because the python one does not fail
- a major issue is that it silently fails, just not setting attributes and returning

agroszer:
fixes #60

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -643,7 +643,10 @@ pickle___setstate__(PyObject *self, PyObject *state)
             for ( i=0; i<len; ++i ) {
                 PyObject *item = PySequence_GetItem(items, i);
                 if (item == NULL)
+                {
+                    Py_DECREF(items);
                     return NULL;
+                }
                 d_key = PyTuple_GetItem(item, 0);
                 if (d_key == NULL)
                 {
@@ -654,7 +657,6 @@ pickle___setstate__(PyObject *self, PyObject *state)
                 d_value = PyTuple_GetItem(item, 1);
                 if (d_value == NULL)
                 {
-                    Py_DECREF(d_key);
                     Py_DECREF(item);
                     Py_DECREF(items);
                     return NULL;
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -588,8 +588,10 @@ pickle___setstate__(PyObject *self, PyObject *state)
     if (state != Py_None)
     {
         PyObject **dict;
+        PyObject *items;
         PyObject *d_key, *d_value;
         Py_ssize_t i;
+        int len;
 
         dict = _PyObject_GetDictPtr(self);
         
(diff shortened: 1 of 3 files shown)


persistent/cPersistence.c
```diff
@@ -609,17 +611,41 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
+                d_key = PyTuple_GetItem(item, 0);
+                d_value = PyTuple_GetItem(item, 1);
+
+                if (NATIVE_CHECK_EXACT(d_key)) {
+                    Py_INCREF(d_key);
+                    INTERN_INPLACE(&d_key);
+                    Py_DECREF(d_key);
+                }
+                if (PyObject_SetItem(*dict, d_key, d_value) < 0)
+                    return NULL;
```
mgedmin: I'm not familiar with the CPython C API, so to check for possible refcounting errors I ran the test suite in a loop (zope-testrunner --repeat 100) and saw no memory usage increase.

IOW, LGTM!

persistent/cPersistence.c
```diff
@@ -609,17 +611,41 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
+            }
+        }
+        else
+        {
+            /* can happen that not a built-in dict is passed as state
+               fall back to iterating over items, instead of silently
+               failing with PyDict_Next */
+            items = PyMapping_Items(state);
```
stephan-hof: Please check for NULL. Which means an exception has been raised.
Take care to call Py_DECREF on `items` when they are not used. Otherwise memory leak.

persistent/cPersistence.c
```diff
@@ -609,17 +611,41 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
+            len = PySequence_Size(items);
```
stephan-hof: I think its better to check `-1` here which signals an error. If that is the case return NULL (and call decref on items).
This re raises the exception from `PySequence_Size` to the caller.

persistent/cPersistence.c
```diff
@@ -609,17 +611,41 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
+            for ( i=0; i<len; ++i ) {
+                PyObject *item = PySequence_GetItem(items, i);
```
stephan-hof: Call Py_DECREF on `item` when not used anymore. Otherwise memory leak.
Also check for NULL in case of errors.

persistent/cPersistence.c
```diff
@@ -609,17 +611,41 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
+                d_key = PyTuple_GetItem(item, 0);
+                d_value = PyTuple_GetItem(item, 1);
+
```
stephan-hof: Check `d_key` and `d_value` for NULL (for example if item is not a tuple) and reraise it to the caller by returning NULL.
Take care to decref `item` and `items` in that case

persistent/cPersistence.c
stephan-hof: In this error case decref `item` and `items` otherwise they are leaking.

mgedmin: (In retrospect, 100 iterations may not be enough to expose a sufficiently small leak, if I'm just looking at the overall memory usage in megabytes.)

agroszer:
@stephan-hof does it look good now?

persistent/cPersistence.c
```diff
@@ -632,20 +632,47 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
             len = PySequence_Size(items);
+            if (len < 0)
+            {
+                Py_DECREF(items);
+                return NULL;
+            }
             for ( i=0; i<len; ++i ) {
                 PyObject *item = PySequence_GetItem(items, i);
+                if (item == NULL)
+                    return NULL;
```
stephan-hof: Please decref `items` otherwise they leak.

persistent/cPersistence.c
```diff
@@ -632,20 +632,47 @@ pickle___setstate__(PyObject *self, PyObject *state)
...
+                if (d_key == NULL)
+                {
+                    Py_DECREF(item);
+                    Py_DECREF(items);
+                    return NULL;
+                }
                 d_value = PyTuple_GetItem(item, 1);
+                if (d_value == NULL)
+                {
+                    Py_DECREF(d_key);
```
stephan-hof: No need to decref `d_key` here, because `PyTuple_GetItem` returns a borrowed reference. Not a new one.

agroszer:
@jimfulton or @tseaver release please

PR #62, opened 2017-03-20
Avoid raising a SystemError when clearing slots if setstate() failed.

jamadden:
PR #52 introduced a code path to `ghostify` that calls PyErr_Clear() with the intent to avoid propagating `AttributeErrors` for slots.

However, if there is an error (like a POSKeyError) raised by jar.setstate(), then `unghostify` will call `ghostify` with an error pending. If the object had slots that weren't set and the `AttributeError` was cleared, so was the pending error from setstate. So when `ghostify` returned NULL that got propagated up to the interpreter which finds no exception and so raises `SystemError: error
return without exception set`.

This commit makes `unghostify` save and restore the exception state around the call to `PyErr_Clear` so the *real* exception is raised, rather than the baffling and unhelpful SystemError.

### DIFF
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -16,6 +16,9 @@ import unittest
 
 import platform
 import sys
+
+from persistent._compat import copy_reg
+
 py_impl = getattr(platform, 'python_implementation', lambda: None)
 _is_pypy3 = py_impl() == 'PyPy' and sys.version_info[0] > 2
 _is_jython = py_impl() == 'Jython'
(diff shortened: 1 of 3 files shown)


persistent/tests/test_persistence.py
```diff
@@ -68,7 +71,7 @@ def register(self, obj):
         jar._cache = self._makeCache(jar)
         return jar
 
-    def _makeBrokenJar(self):
+    def _makeBrokenJar(self, exc=NotImplementedError):
```
tseaver: This extra indirection seems like YAGNI.

jamadden: OK. I was imagining that `NotImplementedError` might be caught and handled specially somewhere so I wanted to be sure to have something that would definitely propagate. But that is hypothetical, I admit.

jamadden: Changed.

jamadden:
@tseaver Could we get a merge and release please?

tseaver:
https://pypi.python.org/pypi/persistent/4.2.4

jamadden:
Thank you!

PR #70, opened 2018-07-30
Change cPickleCache to use PER_TypeCheck

### LINKED ISSUE #69: cPickleCache uses sizeof(X) when it means isinstance(X,persistent)
From cPickleCache.c's [`cc_add` function](https://github.com/zopefoundation/persistent/blob/4d2dcc9352faa05f0ec95ddefc4ebc2537373b4a/persistent/cPickleCache.c#L1037):
```c
    else if (v->ob_type->tp_basicsize < sizeof(cPersistentObject))
    {
        /* If it's not an instance of a persistent class, (ie Python
            classes that derive from persistent.Persistent, BTrees,
            etc), report an error.
            TODO:  checking sizeof() seems a poor test.
        */
        PyErr_SetString(PyExc_TypeError,
                        "Cache values must be persistent objects.");
        return -1;
    }
```

I think I agree with the code's own comment that 'checking sizeof() seems a poor test'. Why does it do that instead of using [`PyObject_IsInstance`](https://docs.python.org/2/c-api/object.html#c.PyObject_IsInstance) (which is what the Python implementation of PickleCache does)? Could that check be changed?

(Ref https://github.com/zopefoundation/Persistence/pull/9)

jamadden:
Instead of testing object sizes.

This matches what the pure Python implementation does and is a stronger test that the object really is compatible with the cache. Previously, an object could potentially include ``cPersistent_HEAD``
and *not* set ``tp_base`` to ``cPersistenceCAPI->pertype`` and still be eligible for the pickle cache; that is no longer the case.

This resolves several compiler warnings:
```
persistent/cPickleCache.c:521:43: warning: comparison of integers of different signs: 'Py_ssize_t' (aka 'long') and 'unsigned long' [-Wsign-compare]
                (v->ob_type->tp_basicsize >= sizeof(cPersistentObject))
                 ~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~
persistent/cPickleCache.c:709:39: warning: comparison of integers of different signs: 'Py_ssize_t' (aka 'long') and 'unsigned long' [-Wsign-compare]
    else if (v->ob_type->tp_basicsize < sizeof(cPersistentObject))
```
Fixes #69.

I ran the BTree tests with this version of persistent installed and they passed, as did the tests for zope.container. Those were the two projects I could think of that have persistent subclasses in C.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,12 +1,20 @@
 ``persistent`` Changelog
 ========================
 
-4.2.5 (unreleased)
+4.3.0 (unreleased)
 ------------------
 
 - Fix the possibility of a rare crash in the C extension when
   deallocating items. See https://github.com/zopefoundation/persistent/issues/66
 
+- Change cPickleCache's comparison of object sizes to determine
+  whether an object can go in the cache to use ``PyObject_TypeCheck()``.
+  This matches what the pure Python implementation does and is a
+  stronger test that the object really is compatible with the cache.
+  Previously, an object could potentially include ``cPersistent_HEAD``
+  and *not* set ``tp_base`` to ``cPersistenceCAPI->pertype`` and still
+  be eligible for the pickle cache; that is no longer the case. See
+  `issue 69 <https://github.com/zopefoundation/persistent/issues/69>`_.
 
 4.2.4.2 (2017-04-23)
 --------------------
--- a/persistent/cPickleCache.c
+++ b/persistent/cPickleCache.c
@@ -99,11 +99,6 @@ static char cPickleCache_doc_string[] =
 #include <stddef.h>
 #undef Py_FindMethod
 
-/* Python 2.4 backward compat */
-#if PY_MAJOR_VERSION <= 2 && PY_MINOR_VERSION < 5
-#define Py_ssize_t int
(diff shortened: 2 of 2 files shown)


CHANGES.rst
```diff
@@ -1,12 +1,20 @@
 ``persistent`` Changelog
 ========================
 
-4.2.5 (unreleased)
+4.3.0 (unreleased)
```
tseaver: Is this a feature bump (seems like a bugfix to me).

jamadden: I went with a minor number bump because it's a C-programmer-visible change in semantics. It's subtle, but it's there, and I think there is a small chance it could break some previously working code.

But if it's a subtle enough change you'd prefer calling it a bug fix, I'm cool with that too :)

tseaver: I think I voted on #69 that anybody who can detect the change externally already Deserves to Lose(TM), but we can leave the bump in place if you want to be nice. :)

jamadden: I'm feeling generous today :)

jamadden:
Thanks!

PR #72, opened 2018-07-30
Update url in setup.py

### LINKED ISSUE #71: The setup.py `url` is broken
Currently it points to "http://www.zope.org/Products/ZODB", which is a 404.

Most zopefoundation projects use their github page as the url these days, which has the nice side-effect of putting some extra information on PyPI.

jamadden:
Fixes #71

Also fix the lint warnings in setup.py: consistent use of continued indentation and trailing commas, no spaces around = in keyword arguments

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ from setuptools import Extension
 from setuptools import find_packages
 from setuptools import setup
 
-version = '4.2.5.dev0'
+version = '4.3.0.dev0'
 
 here = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -20,7 +20,7 @@
 from setuptools import find_packages
 from setuptools import setup
 
-version = '4.2.5.dev0'
+version = '4.3.0.dev0'
```
tseaver: Rebase to pick up that this is done already in #70?

jamadden: I missed changing setup.py in that PR, I just hit CHANGES.rst.

tseaver: That's why I usually avoid doing either of those changes until just before making the release.

jamadden:
Thank you. I intend to release to PyPI now if there are no objections.

PR #79, opened 2018-07-31
Build CFFI modules the modern way.

### LINKED ISSUE #43: Distribute the CFFI ring implementation in the wheels?
This is mostly a problem for PyPy users on Windows (who probably don't have compilers).

We should be able to do this with this snippet in `setup.py` (having made sure to install CFFI and zope.interface first; the zope.interface requirement could be dropped with some refactoring):

```
try:
    import cffi
    import persistent.ring
except ImportError:
    pass
else:
    ext_modules.append(persistent.ring.ffi.verifier.get_extension())
```

That should work back to PyPy 2.5.1 or so, which is the version on Travis.

Now the second half of this is that we don't distribute PyPy wheels, so to meet the stated purpose we'd also need to do that (at least on Windows)

### LINKED ISSUE #75: Use the modern way to build CFFI modules
We're using an old, semi-deprecated way that was required by ancient versions of CFFI as distributed in ancient versions of PyPy. Instead of building the module at install time, it builds it at import time. This means it doesn't get in the wheels (see #43).

jamadden:
Closes #75. 
Closes #43.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,11 +6,14 @@
 
 - Add support for Python 3.7 and drop support for Python 3.3.
 
-- Build the CFFI modules (used on PyPy or when PURE_PYTHON is set) at
-  installation or wheel building time when CFFI is available. This
-  replaces the deprecated way of building them at import time. If
-  binary wheels are distributed, it eliminates the need to have a
-  functioning C compiler to use PyPy. See `issue 75
+- Build the CFFI modules (used on PyPy or when PURE_PYTHON is set) `at
+  installation or wheel building time
+  <https://cffi.readthedocs.io/en/latest/cdef.html#ffibuilder-set-source-preparing-out-of-line-modules>`_
+  when CFFI is available. This replaces `the deprecated way
+  <https://cffi.readthedocs.io/en/latest/overview.html#abi-versus-api>`_
+  of building them at import time. If binary wheels are distributed,
+  it eliminates the need to have a functioning C compiler to use PyPy.
+  See `issue 75
   <https://github.com/zopefoundation/persistent/issues/75>`_.
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,13 @@
 
 - Add support for Python 3.7 and drop support for Python 3.3.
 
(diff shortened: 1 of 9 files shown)


CHANGES.rst
```diff
@@ -6,6 +6,13 @@
 
 - Add support for Python 3.7 and drop support for Python 3.3.
 
+- Build the CFFI modules (used on PyPy or when PURE_PYTHON is set) at
+  installation or wheel building time when CFFI is available. This
+  replaces the deprecated way of building them at import time. If
+  binary wheels are distributed, it eliminates the need to have a
+  functioning C compiler to use PyPy. See `issue 75
+  <https://github.com/zopefoundation/persistent/issues/75>`_.
```
tseaver: My CFFIfu is weak: can you add a link to docs about how the "modern way" is supposed to work?

jamadden: Sure will. It's https://cffi.readthedocs.io/en/latest/cdef.html#ffibuilder-set-source-preparing-out-of-line-modules for "API + out of line"

What we *were* doing was "API + in line", [which is deprecated](https://cffi.readthedocs.io/en/latest/overview.html#abi-versus-api):

> Note that the “API level + in-line” mode combination exists but is long deprecated. It used to be done with lib = ffi.verify("C header"). The out-of-line variant with set_source("modname", "C header") is preferred.

tseaver: Wonderful, thanks!

setup.py
```diff
@@ -110,6 +110,7 @@ def _read_file(filename):
       include_package_data=True,
       zip_safe=False,
       ext_modules=ext_modules,
+      cffi_modules=['persistent/_ring_build.py:ffi'],
```
tseaver: Does this bit need a `setup_requires=["cffi>=1.0.0"]`?  That seems to be in the [example](https://cffi.readthedocs.io/en/latest/cdef.html#preparing-and-distributing-modules), anyway.

jamadden: No, that's not necessary. setuptools ignores unknown keyword arguments. 

It does mean that we need to make sure we have cffi installed before we build the wheels to distribute, though, if we want the cffi module to be built. This PR does that.

jamadden:
Thank you!

Issue #76, opened 2018-07-31
Enable coveralls and reach 100% coverage


jamadden:

jamadden:
Closed in #80.

PR #80, opened 2018-07-31
Reach 100% coverage

jamadden:
In tox only for now, coveralls coming in a followup PR or commit.

There are also two minor pure-python bug fixes I discovered.

Addresses #76.

### DIFF
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -99,8 +99,7 @@ class _Persistent_Base(object):
         return jar
 
     def _makeOneWithJar(self, klass=None, broken_jar=False):
-        from persistent.timestamp import _makeOctets
-        OID = _makeOctets('\x01' * 8)
+        OID = b'\x01' * 8
         if klass is not None:
             inst = klass()
         else:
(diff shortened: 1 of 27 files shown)


persistent/tests/test_persistence.py
```diff
@@ -1743,17 +1736,17 @@ def test_accessed_with_jar_and_oid_but_not_in_cache(self):
...
         c1._p_oid = KEY
         c1._p_jar = jar
-        orig_mru = jar._cache.mru
+
         def mru(oid):
             # Mimic what the real cache does
             if oid not in jar._cache._mru:
                 raise KeyError(oid)
-            orig_mru(oid)
+            raise AssertionError("Shold never get here")
```
mgedmin: Shold ->  Should

jamadden:
Thanks. For the most part it's very small changes, just quite a few of them. The most potentially disruptive change is [in init.py](
https://github.com/zopefoundation/persistent/pull/80/commits/97ef73c620db6c60ddb889b5b07852929437ab1f), which I simplified substantially, including taking out an attempted import of `persistent.TimeStamp` (note case) which should always fail. As far as I can see that's a rename that happened in the distant past.

I'll add this to CHANGES.

PR #83, opened 2018-08-01
Make the C and Python TimeStamp round the same way

### LINKED ISSUE #41: pyTimeStamp and TimeStamp have different sub-second precision
I don't think this is correct:

```
>>> from persistent.timestamp import TimeStamp, pyTimeStamp
>>> ts1 = TimeStamp(2001, 2, 3, 4, 5, 6.123456789)
>>> ts2 = pyTimeStamp(2001, 2, 3, 4, 5, 6.123456789)
>>> ts1 == ts2
True
>>> ts1.second() == ts2.second()
False
>>> ts1.second()
6.1234567780047655
>>> ts2.second()
6.123456789
```

jamadden:
Fixes #41.

The C version now always rounds its seconds value, even when its decoding from a value with lots of noise, as when initially created from 6.123456789. Previously it would return 6.1234567780047655; now it returns 6.123457 which is what it would return after being round-tripped through raw().

The Python version also does this, by (expensively) unpacking the raw value it just computed.

These could both probably be better optimized.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,15 @@
 - Remove some internal compatibility shims that are no longer
   necessary. See `PR 82 <https://github.com/zopefoundation/persistent/pull/82>`_.
 
+- Make the return value of ``TimeStamp.second()`` consistent across C
+  and Python implementations when the ``TimeStamp`` was created from 6
+  arguments with floating point seconds. Also make it match across
+  trips through ``TimeStamp.raw()``. Previously, the C version could
+  initially have erroneous rounding and too much false precision,
+  while the Python version could have too much precision. The raw/repr
+  values have not changed. See `issue 41
+  <https://github.com/zopefoundation/persistent/issues/41>`_.
+
 4.3.0 (2018-07-30)
 ------------------
 
--- a/persistent/tests/test_timestamp.py
+++ b/persistent/tests/test_timestamp.py
@@ -17,6 +17,8 @@ import sys
 MAX_32_BITS = 2 ** 31 - 1
 MAX_64_BITS = 2 ** 63 - 1
 
+import persistent.timestamp
+
 class Test__UTC(unittest.TestCase):
 
     def _getTargetClass(self):
@@ -202,7 +204,8 @@ class TimeStampTests(pyTimeStampTests):
         from persistent.timestamp import TimeStamp
         return TimeStamp
 
-
+@unittest.skipIf(persistent.timestamp.CTimeStamp is None,
+                 "CTimeStamp not available")
 class PyAndCComparisonTests(unittest.TestCase):
     """
     Compares C and Python implementations.
@@ -254,7 +257,6 @@ class PyAndCComparisonTests(unittest.TestCase):
 
     def test_equal(self):
         c, py = self._make_C_and_Py(*self.now_ts_args)
-
         self.assertEqual(c, py)
 
     def test_hash_equal(self):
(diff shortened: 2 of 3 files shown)


jamadden:
Another option, instead of having the C code round each time it retrieves `TimeStamp_sec`, would be to have both versions do the rounding before creating the raw value. That loses precision in the raw value, though, potentially leading to more collisions.

jamadden:
Argh. Apparently MSVC 9 doesn’t have the standard `round` function, at least not by default.

CHANGES.rst
```diff
@@ -31,6 +31,14 @@
 - Remove some internal compatibility shims that are no longer
   necessary. See `PR 82 <https://github.com/zopefoundation/persistent/pull/82>`_.
 
+- The precision of ``TimeStamp`` objects created from 6 arguments with
```
jamadden: I will clarify that this refers to the return of the `second()` function: the internal precision used for the `raw()` is unchanged.

mgedmin:
>  I see you already have a unit test for this case (6.5), and I wonder why it passes?

That's because I'm an idiot who forgot that it's not `round(...)` that applies here but `round(..., 6)`.

jamadden:
> I think that C and Python versions should be consistent between themselves, but I don't think round(..., 6) is appropriate.

That's essentially what happens in the raw bytes format, and why a trip through `raw()` and back previously appeared to lose that precision in the Python version. The C version is always derived from the raw version, which is why it showed that loss immediately. I did have an implementation of the Python version that was basically `self._elements = parseRaw(makeRaw(*elements))` but that seemed needlessly expensive, though that may be premature optimization. Would you prefer that? Or can you suggest a different approach?

mgedmin:
> That's essentially what happens in the raw bytes format

Ah, does the raw bytes version store an integer number of microseconds?  I should've refreshed my memory before attempting to review code.

> self._elements = parseRaw(makeRaw(*elements))

I do actually prefer that (with a comment explaining that this is done to discard extra precision that wouldn't be preserved in the raw format).

(TBH the thing I would prefer _most_ is for somebody smarter than me to review this PR.)

jamadden:
> 

It does, yes. 

Python:
```python
def _makeRaw(year, month, day, hour, minute, second):
    a = (((year - 1900) * 12 + month - 1) * 31 + day - 1)
    a = (a * 24 + hour) * 60 + minute
    b = int(second / _SCONV) # Don't round() this; the C version does simple truncation
    return struct.pack('>II', a, b)
```

C:
```C
PyObject *
TimeStamp_FromDate(int year, int month, int day, int hour, int min,
           double sec)
{
    TimeStamp *ts = NULL;
    //...
    /// ts->data is unsigned char[]
    sec /= SCONV;
    v = (unsigned int)sec;
    ts->data[4] = v / 16777216;
    ts->data[5] = (v % 16777216) / 65536;
    ts->data[6] = (v % 65536) / 256;
    ts->data[7] = v % 256;
    //...
}
```

> I do actually prefer that (with a comment explaining that this is done to discard extra precision that wouldn't be preserved in the raw format).

I can go back to that in the constructor. That's slightly closer parity with C. But I think I would still need a `round()` call in the `seconds()` accessor: [the C version needed that](https://github.com/zopefoundation/persistent/pull/83/files#diff-6a6e068aef7fec9e0c363cbce457b2a9R212).  The problem, I think, is that the storage format is lossy, and without the `round()` sometimes the results don't make sense (they're not clearly derived from the input).

jamadden:
>  But I think I would still need a round() call in the seconds() accessor: the C version needed that.

No, the C version needed that precisely because the constructor doesn't round-trip through raw().

FWIW, this is a place where the compiler flag `-ffast-math` breaks things. The tests weren't passing for me this morning (specifically, 6.5555 was producing 6.555499999999999 in the C version), and it turned out to be because I compiled with `-ffast-math`.

jamadden:
>> But I think I would still need a round() call in the seconds() accessor: the C version needed that.

> No, the C version needed that precisely because the constructor doesn't round-trip through raw().

Once again, I am wrong. The C version always constructs the double value from its raw bytes, which don't change across trips through raw():
```C
    unsigned int v;
    double sec;

    v = (self->data[4] * 16777216 + self->data[5] * 65536
     + self->data[6] * 256 + self->data[7]);
    sec = SCONV * (double)v;
```

The Python version explicitly calls `round()`  when it parses raw bytes, so that's why the C version needs to do so as well. They either both need to or both shouldn't. If they both *do not* call `round` at all, then there is a single test failure in the Python implementation: `TimeStamp(2011, 2, 16, 14, 37, 22.80544).timeTime()` produces 1297867042.805439 instead of 1297867042.80544. That can be fixed with a change to `_makeUTC`.

jamadden:
ping :)

jamadden:
I backed out the textual change to the C source file, so this PR is pretty simple now, adding a (previously failing) test and updating the Python implementation to fix it.

jamadden:
Thanks!

Issue #11, opened 2014-07-16
RFC: Better default __repr__ for persistent objects


jimfulton:
Persistent objects inherit their default `__repr__` from object.  This shows the object's address:

```
<somemodule.SomeClass object at 0x102a3d460>
```

I propose that when a persistent object is stored in a database, the default `__repr__` should be changed to:

```
<somemodule.SomeClass object OID_REPR in DB_REPR>
```

In this way, objects whose `__reprs__` appear in logs or errors, can, at least potentially, looked up for inspection.

In addition, I propose defining a new method `_p_repr`.

If defined, it will be used instead of the default `__repr__`.  If `_p_repr__` fails because it's dabase connection is closes, then fall back to the default.   Due to the desire to loosen the dependency of `Persistent` on ZODB, some additional abstract methods might be needed to determine if this is the source of the failure. An alternative would be to add something to the default `__repr__` to indicate that it was called because the `_p_repr` failed.

PR #84, opened 2018-08-02
Include OID and jar in default repr of Persistent; let subclasses implement _p_repr for exception safety

jamadden:
As specified in #11. See the tests for what the output looks like; in the usual case it will be something like `"<persistent.Persistent object at 0xdeadbeef oid b'12345678' in <SomeJar>>"`

I would appreciate a thorough check of the C code, I'm not a C API expert.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@
 4.3.1 (unreleased)
 ------------------
 
+- Change the default representation of ``Persistent`` objects to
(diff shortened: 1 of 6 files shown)


jamadden:
ping :)

jamadden:
Hopefully if they use ellipsis for the address, this won’t break them since it’s still first. Something more sophisticated would still fail though. 

Thanks for the review!

Issue #24, opened 2015-05-19
Should `_p_oid` allow arbitrary types?


tseaver:
The docstring for `IPersistent._p_oid` claims that it must be bytes.  Historically, the C implementation has allowed any type (although values which are not length 8 bytes blow up if used with a standard pickle cache).

See:  https://github.com/zopefoundation/persistent/pull/20#discussion-diff-28083537 for pros and cons.

jamadden:
There's additional discussion on that same PR [here](https://github.com/zopefoundation/persistent/pull/20#discussion_r30594848).

jamadden:
I just discovered another inconsistency in https://github.com/NextThought/ZODB/commit/5d38a086be4bcf20558d8bcec33cd95d065a7082.

It turns out that the cPickleCache implementation only enforces the 8-length byte constraint in the `__setitem__` implementation. If you use `new_ghost`, it accepts any arbitrary value for the oid. The Python implementation enforces it both places.

ZODB initially creates and populates the picklecache using `new_ghost` ([Connection.get](https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/Connection.py#L264)), so in practice it might seem like if you're using the cPickleCache the restriction is easily bypassed. However, once you commit the transaction, if you modified any of those objects, *then* you'll find that it calls `__setitem__` and things blow up with a TypeError simply because cPickleCache checks the type of the key before checking to see if the key already exists.

strichter:
I'll note that when I use a non-bytes _p_oid, I do not use pickle at all. Whenever pickle is used, you might as well use the ZODB facilities, because it takes care of most things you would ever want.

For me I specifically went down to the persistent layer for pjpersist because it does not deal with serialization and only provides the messaging framework.

tseaver:
I'll close this with the note that the consensus answer is "No, OIDs do not have to be bytes of length 8" for implementations which use jars other than those supplied by ZODB.

PR #95, opened 2018-10-04
Always represent oids as hex

Rotonen:
We hit an issue in Plone testing where having bad sequences of bytes in Zope object `__repr__`s could actually take down a whole test run as it took down lxml / libxml at the write time of a JUnit test report file.

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -1423,6 +1423,8 @@ Per_repr(cPersistentObject *self)
     PyObject *prepr = NULL;
     PyObject *prepr_exc_str = NULL;
 
+    PyObject *bytes_hex(PyObject *bytes);
+
     PyObject *oid_str = NULL;
     PyObject *jar_str = NULL;
     PyObject *result = NULL;
@@ -1446,7 +1448,7 @@ Per_repr(cPersistentObject *self)
         prepr_exc_str = PyUnicode_FromString("");
     }
 
-    oid_str = repr_helper(self->oid, " oid %R");
+    oid_str = repr_helper(bytes_hex(self->oid), " oid 0x%s");
     if (!oid_str)
         goto cleanup;
 
--- a/persistent/persistence.py
+++ b/persistent/persistence.py
@@ -574,7 +574,16 @@ class Persistent(object):
 
         if oid is not None:
             try:
-                oid_str = ' oid %r' % (oid,)
+                if isinstance(oid, bytes):
+                    # Python < 3.2
+                    if hasattr(oid, 'encode'):
+                        oid = ''.join(('0x', oid.encode('hex'), ))
+                    # Python >= 3.2
+                    else:
+                        oid = hex(int.from_bytes(oid, byteorder='big'))
+                    oid_str = ' oid %s' % (oid, )
+                else:
+                    oid_str = ' oid %r' % (oid, )
             except Exception as e:
                 oid_str = ' oid %r' % (e,)
 
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -1702,11 +1702,11 @@ class _Persistent_Base(object):
         r = r.replace('persistent.persistence.Persistent', 'persistent.Persistent')
         r = r.replace("persistent.tests.test_persistence.", '')
         # addresses
-        r = re.sub(r'0x[0-9a-fA-F]*', '0xdeadbeef', r)
+        r = re.sub(r'at 0x[0-9a-fA-F]*', 'at 0xdeadbeef', r)
         # Python 3.7 removed the trailing , in exception reprs
         r = r.replace("',)", "')")
         # Python 2 doesn't have a leading b prefix for byte literals
-        r = r.replace("oid '", "oid b'")
+        # r = r.replace("oid '", "oid b'")
         return r
 
     def _normalized_repr(self, o):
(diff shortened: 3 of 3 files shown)


jamadden:
This layer doesn't strictly enforce that `_p_oid` is required to be a bytestring with len 8. Arbitrary objects are allowed. See https://github.com/zopefoundation/persistent/issues/24

What was the error that caused problems? Is it just that a something created a malformed XML file?

Rotonen:
We've been having issues with our doctests as well as there can be `'`, `"` or any `\` beginning sequence within an oid, which is "just an int" or "just a bunch of bytes", getting interpreted as a string.

At least the arbitrarily long sequence of bytes should be read as an int and represented as a hex, which is what I was going for, but jumped the gun a bit on the PR in the hopes of catching more attention (and also giving us the chance to test our stuff against the branch here).

jensens:
What about to check if it is bytes (of any length) and convert to hex, otherwise use the objects repr?

Rotonen:
I think I'm inching towards something more acceptable. Would anyone know how the win32 linking works? I'm not spotting what actually generates the signature manifests.

jamadden:
Where are you expecting the function `bytes_hex` to come from? I don't find it [to be a documented C-API function](https://docs.python.org/3/search.html?q=bytes_hex&check_keywords=yes&area=default) and I don't find it defined in this code base. I can *only* find it defined in CPython's byteobject.c as `static PyObject * bytes_hex()`. static functions are not linkable or usable outside their defining file.

Rotonen:
So my question is actually the opposite, why do clang and gcc even resolve the symbol on that one. The root cause being thinking in Python and not realising only the headers concatenate into your current context in C.

So, it seems I'll need to actually dive in and use the API proper here and potentially use the version macros for divergent implementations where need be, mirroring what is done on the Python side.

Before I do that, @jamadden, is the approach taken palatable?

jamadden:
> So my question is actually the opposite, why do clang and gcc even resolve the symbol on that one. 

They don't. But we're building a dynamic link library with options that say "find the rest of the symbols at runtime" (with clang on OS X, that's ` -undefined dynamic_lookup`, on other GCC platforms that might be implicit?). So runtime comes along and the dynamic linker is asked to resolve this symbol and it can't. Linking failures at the shared library level are propagated to Python as `ImportError`, and an `ImportError` of `cPersistence` [triggers us to just use the pure-Python implementation](https://github.com/zopefoundation/persistent/blob/master/persistent/__init__.py#L43):
```pycon
>>> import persistent
>>> import persistent.cPersistence
ImportError: dlopen(/persistent/persistent/cPersistence.so, 2): Symbol not found: _bytes_hex
  Referenced from: /persistent/persistent/cPersistence.so
  Expected in: flat namespace
 in //persistent/persistent/cPersistence.so
>>> persistent.Persistent
<class 'persistent.persistence.Persistent'>
>>> import persistent.cPersistence
```

jamadden:
> Before I do that, @jamadden, is the approach taken palatable?

I'm sorry, I still don't understand the problem that needs to be solved. 

You write:

> We've been having issues with our doctests as well as there can be ', " or any \ beginning sequence within an oid, which is "just an int" or "just a bunch of bytes", getting interpreted as a string.

I don't understand why this is a problem. `repr` of a bytestring has a well-defined format, even on Python 2, and it's always printable. Can't `...` in the doctest deal with that?
```pycon
>>> from ZODB.utils import p64
>>> for _ in range(10):
    ...:     print(repr(p64(random.randint(1, 999999999))))
"\x00\x00\x00\x00'&\xfd\x1d"
'\x00\x00\x00\x009\xad\x8a\x01'
'\x00\x00\x00\x005&\xcd\xc0'
'\x00\x00\x00\x00\x04\xa5g\xb6'
'\x00\x00\x00\x00\x02\x9e\x93\xbf'
"\x00\x00\x00\x00'\xf7R\x16"
'\x00\x00\x00\x00\x15\xebw\xa0'
'\x00\x00\x00\x00(\x82\xab\xb7'
'\x00\x00\x00\x00-\x9a\x13\xcd'
'\x00\x00\x00\x0009\xdd\x04'
```

That said, there is precedent for hexlifying 8-byte strings. ZODB's [POSKeyError](https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/POSException.py#L55) does this with an [`oid_repr` function](https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/utils.py#L145). But it's a moderately complex function that doesn't look fun to replicate safely in C.

So to answer the question, I guess I'm a slight -0, between not understanding the problem and being afraid of the complexity, but I'm not the one offering to do the work.

jamadden:
> Linking failures at the shared library level are propagated to Python as ImportError, and an ImportError of cPersistence triggers us to just use the pure-Python implementation:

Hmm, the fact that the Travis builds all passed suggests that we're missing a test assertion. As far as these tests go, if it's not `PURE_PYTHON` (in the environment or because we're on PyPy) we should be able to assert that `persistent._cPersistence` is not None (or that an import of it doesn't fail), otherwise we can silently introduce C extension problems. Thankfully the Appveyor builds caught the issue, but locally it took me a few minutes debugging to understand why.

jamadden:
I found https://github.com/plone/plone.testing/pull/53/ which makes the problem more clear to me, so I will go up to a +0 😄

Rotonen:
Thank you for digging in, makes sense. For that I'll actually have to tackle the C side properly. I'll also amend the C implementation testing to assert the implementation is actually used.

Gnarly, but still a lower hanging fruit to ensure ASCII at the source for me than to ensure various tooling to work proper with arbitrary inputs. The doctest one is a relatively easy one, but feeding 'bad strings' into other tooling, like JUnit XML writers, is resulting in crashes.

jamadden:
I think I see how to do this generically in C. Working on a new PR.

PR #98, opened 2018-10-19
Always format 8-byte oids as ints in hexadecimal.

jamadden:
Fixes #95

Built on #97 as they touch the same code.

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -16,7 +16,6 @@ static char cPersistence_doc_string[] =
   "\n"
   "$Id$\n";
 
-#include <stdint.h>
 #include "cPersistence.h"
 #include "structmember.h"
 
@@ -24,6 +23,16 @@ struct ccobject_head_struct
 {
     CACHE_HEAD
 };
+
+/*
+  The compiler on Windows used for Python 2.7 doesn't include
+  stdint.h.
+*/
+#if !defined(PY3K) && defined(_WIN32)
+typedef unsigned long long uint64_t;
+#else
+#include <stdint.h>
+#endif
 
 /* These two objects are initialized when the module is loaded */
 static PyObject *TimeStamp, *py_simple_new;
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -16,6 +16,7 @@ static char cPersistence_doc_string[] =
   "\n"
   "$Id$\n";
 
+#include <stdint.h>
 #include "cPersistence.h"
 #include "structmember.h"
 
@@ -1430,7 +1431,8 @@ Per_repr(cPersistentObject *self)
     PyObject *result = NULL;
 
     unsigned char* oid_bytes;
-    unsigned long long oid_value;
+    char buf[20];
+    uint64_t oid_value;
 
     prepr = PyObject_GetAttrString((PyObject*)Py_TYPE(self), "_p_repr");
     if (prepr)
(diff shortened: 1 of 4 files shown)


persistent/cPersistence.c
```diff
@@ -1446,23 +1451,50 @@ Per_repr(cPersistentObject *self)
...
+        oid_bytes = (unsigned char*)PyBytes_AS_STRING(self->oid);
+        oid_value = ((unsigned long long)oid_bytes[0] << 56)
+            | ((unsigned long long)oid_bytes[1] << 48)
+            | ((unsigned long long)oid_bytes[2] << 40)
+            | ((unsigned long long)oid_bytes[3] << 32)
+            | ((unsigned long long)oid_bytes[4] << 24)
+            | ((unsigned long long)oid_bytes[5] << 16)
+            | ((unsigned long long)oid_bytes[6] << 8)
+            | ((unsigned long long)oid_bytes[7]);
+        oid_str = PyUnicode_FromFormat(" oid 0x%x", oid_value);
```
mgedmin: [The documentation](https://docs.python.org/3/c-api/unicode.html) says %x takes an `int`, not an `unsigned long long`.

There are many platforms where an `int` is shorter than 64 bits.

We should at least have a unit test that uses an OID that doesn't fit in 32 bits.

persistent/cPersistence.c
```diff
@@ -1446,23 +1451,50 @@ Per_repr(cPersistentObject *self)
...
+    if (self->oid && PyBytes_Check(self->oid) && PyBytes_GET_SIZE(self->oid) == 8) {
+        oid_bytes = (unsigned char*)PyBytes_AS_STRING(self->oid);
+        oid_value = ((unsigned long long)oid_bytes[0] << 56)
+            | ((unsigned long long)oid_bytes[1] << 48)
+            | ((unsigned long long)oid_bytes[2] << 40)
+            | ((unsigned long long)oid_bytes[3] << 32)
+            | ((unsigned long long)oid_bytes[4] << 24)
+            | ((unsigned long long)oid_bytes[5] << 16)
+            | ((unsigned long long)oid_bytes[6] << 8)
+            | ((unsigned long long)oid_bytes[7]);
```
mgedmin: Could we use uint64_t instead of unsigned long long?

persistent/tests/test_persistence.py
```diff
@@ -1733,12 +1733,27 @@ def __repr__(self):
...
         result = self._normalized_repr(p)
         self.assertEqual(
             result,
-            "<persistent.Persistent object at 0xdeadbeef oid b'12345678'>")
+            "<persistent.Persistent object at 0xdeadbeef oid " + self._HEX_OID + ">")
+
+    def test_64bit_oid(self):
+        import struct
+        p = self._makeOne()
+        oid_value = 2 << 62
```
mgedmin: 1 << 63 would be clearer imho

jamadden:
Thanks! I don’t consider myself a c api expert either, so I welcome all reviews.

I know that downstream is waiting on this though, and because I feel bad that my implementation of the requested improvements broke things, so if there are no further comments in, say, the next 36 hours, I expect to go ahead and merge and release.

persistent/tests/test_persistence.py
mgedmin: I was looking at this, wondering why 2**62 instead of 2**63, and then I realized << is not ** :D

icemac:
@jamadden +1

icemac:
@jamadden 👍

Rotonen:
@jamadden I owe you a `$BEVERAGE` should our paths ever cross.

jamadden:
4.4.3 is on PyPI.

PR #104, opened 2018-11-11
Improve cache_size and cache_size_bytes functionality.

ramiroluz:
This work was done with @yuseitahara, I am the commiter because I did the final work on my machine.


yuseitahara:
@ramiroluz and I work on zopefoundation/ZODB/issues/231 and this is our working on progress result. We found that cPickleCache is not tested, we added two tests, more tests will come.

PR #105, opened 2018-11-12
Fully test the C implementation of the PickleCache

### LINKED ISSUE #102: is cPickleCache tested?
Looking at the tests in test_picklecache, it looks like they're only testing the Python version.

Am I missing something?

jamadden:
Fix discrepancies between it and the Python implementation:

  - The C implementation now raises ``ValueError`` instead of ``AssertionError`` for certain types of bad inputs.
  - The Python implementation uses the C wording for error messages.
  - The C implementation properly implements ``IPickleCache``; methods  unique to the Python implementation were moved to ``IExtendedPickleCache``.
  - The Python implementation raises ``AttributeError`` if a  persistent class doesn't have a ``p_jar`` attribute.

Fixes #102

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,10 +1,20 @@
 ``persistent`` Changelog
 ========================
 
-4.4.4 (unreleased)
+4.5.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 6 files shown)


jamadden:
Thanks! (FWIW, I also ran the ZODB tests with this without issue.)

PR #110, opened 2019-04-24
Drop support for Python 3.4, add support for 3.8.

icemac:


tox.ini
```diff
@@ -20,20 +20,26 @@ basepython =
...
+    PURE_PYTHON = 1
+
 [testenv:coverage]
 usedevelop = true
 basepython =
     python3.6
 commands =
     coverage run -m zope.testrunner --test-path=.
-    coverage report --fail-under=93
+    coverage report --fail-under=100
```
mgedmin: :100:!

docs/using.rst
```diff
@@ -376,10 +376,10 @@ The ``_p_serial`` attribute is not affected by calling setstate.
 
 .. doctest::
 
-   >>> p._p_serial = "00000012"
+   >>> p._p_serial = b"00000012"
    >>> p.__setstate__(p.__getstate__())
    >>> p._p_serial
-   '00000012'
+   b'00000012'
```
mgedmin: Wow are these doctests never executed on Travis CI?

jamadden: The doctests are currently using sphinx, which makes it somewhat more complicated to write cross-version tests (no normalizers).

tox.ini
```diff
@@ -20,20 +20,26 @@ basepython =
 setenv =
     PURE_PYTHON = 1
 
+[testenv:py37-pure]
+basepython =
+    python3.7
+setenv =
+    PURE_PYTHON = 1
```
mgedmin: Maybe we want to add this to the Travis build matrix as well?

jamadden: Or does the PyPy3 branch effectively cover that?

docs/using.rst
icemac: I don't think so but until this PR they where run on Python 2.7.

mgedmin: Ah, right.  Never mind then.

(But maybe we want to add a build job to run one sphinx-build doctest for the specific Python 3 version we're specifying in tox.ini).

.travis.yml
```diff
@@ -38,6 +34,9 @@ matrix:
         - os: osx
           language: generic
           env: TERRYFY_PYTHON='macpython 3.7.0'
+        - os: osx
+          language: generic
+          env: TERRYFY_PYTHON='macpython 3.8.0a3'
```
jamadden: This [doesn't appear to work](https://travis-ci.org/zopefoundation/persistent/jobs/523996701)

mgedmin:
Heads up: I dropped Python 3.4 support in #111 for reasons, which may have introduced merge conflicts with your PR.

I apologize for the inconvenience!

mgedmin:
Can you please rebase so there are no merge commits in the middle of the PR?

icemac:
My plan was to squash merge this PR to get rid of the back and forth (e. g. with Terryfy and Python 3.8).
Rebasing requires a force push and I already got negative comments for using force pushes in PRs because it makes it hard to follow the changes. I even changed the settings in the [Zope](https://github.com/zopefoundation/Zope) project to only allow squash merges. This means everything can happen on a branch (GitHub allows to merge from the destination branch with one click!) but there will only be one clean commit on master.

What do you think about this?

mgedmin:
Sure, a squashed merge seems fine!

jamadden:
> My plan was to squash merge this PR to get rid of the back and forth (e. g. with Terryfy and Python 3.8).
> Rebasing requires a force push and I already got negative comments for using force pushes in PRs because it makes it hard to follow the changes. I even changed the settings in the Zope project to only allow squash merges. This means everything can happen on a branch (GitHub allows to merge from the destination branch with one click!) but there will only be one clean commit on master.

> What do you think about this?

Speaking for myself, I'm fine with occasional judicious use of squash merges, but I do not like the idea of forcing that as a policy. There can be lots of useful information in commits, even if those commits are later reverted. Knowing *how* a particular state was reached, what was tried and what wasn't, can be even more useful than just knowing the end result.  In an ideal scenario, that information winds up recorded in new comments in the files themselves (requiring new commits), but too often a sequence of commits "do abc" and then "revert abc because it breaks xyz in fascinating way 123" just winds up squashed out as a no-op and the information that abc was tried, it broke xyz because of 123 is just lost.

If a branch has a particularly messy commit history with lots of *useless* thrashing, sure, rebase it and squash some of those individual commits into coherent units as a final step requiring a force push. But don't throw away information.

mgedmin:
> Rebasing requires a force push and I already got negative comments for using force pushes in PRs because it makes it hard to follow the changes. 

FWIW my perspective on this is:

- when a reviewer requests changes, don't rebase, as that will force the reviewer to do a full review of everything again, which is very annoying if the diff is large
- when a review is done, provisional LGTMs have been given, and all that remains is to resolve some merge conflicts, a rebase + forced push is Good

PR #115, opened 2019-07-14
FIX: Clearing a persistent list does not mark it as changed.

lungj:
The inherited `clear()` method in UserList can modify the `PersistentList` without causing the object to be marked as changed. This patch causes the `clear()` method to mark the object as changed.

### DIFF
--- a/persistent/list.py
+++ b/persistent/list.py
@@ -15,16 +15,24 @@
 """Python implementation of persistent list.
 
 $Id$"""
-
+import sys
 import persistent
 from persistent._compat import UserList
 from persistent._compat import PYTHON2
+
+# The slice object you get when you write list[:]
+_SLICE_ALL = slice(None, None, None)
 
 class PersistentList(UserList, persistent.Persistent):
     """A persistent wrapper for list objects.
 
     Mutating instances of this class will cause them to be marked
     as changed and automatically persisted.
+
+    .. versionchanged:: 4.5.2
+       Using the `clear` method, or writing ``del inst[:]`` now only
+       results in marking the instance as changed if it actually removed
+       items.
     """
     __super_setitem = UserList.__setitem__
     __super_delitem = UserList.__delitem__
@@ -37,14 +45,23 @@ class PersistentList(UserList, persistent.Persistent):
     __super_reverse = UserList.reverse
     __super_sort = UserList.sort
     __super_extend = UserList.extend
+    __super_clear = (
+        UserList.clear
+        if hasattr(UserList, 'clear')
+        else lambda inst: inst.__delitem__(slice(None, None, None))
+    )
 
     def __setitem__(self, i, item):
         self.__super_setitem(i, item)
         self._p_changed = 1
 
     def __delitem__(self, i):
+        # If they write del list[:] but we're empty,
+        # no need to mark us changed.
+        needs_changed = i != _SLICE_ALL or bool(self)
         self.__super_delitem(i)
-        self._p_changed = 1
+        if needs_changed:
+            self._p_changed = 1
 
     if PYTHON2:  # pragma: no cover
         __super_setslice = UserList.__setslice__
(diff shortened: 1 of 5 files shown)


jamadden:
Thanks for the pull request! That's a great find.

It's missing just a few small things. It'll need a test case added to `persistent/tests/test_list.py` (and that `# pragma: no cover` removed, I think), and it'll need a change note added to `CHANGES.rst`. Let me know if I can assist with that.

We may also need a contributor agreement to be signed. I'm not 100% sure what the status of that is, though. @tseaver may be able to help out.

lungj:
Wasn't sure whether to change the tests file to import `PYTHON2` globally in order to only run `test_clear` if not running with Python 2. Made the test run always but only execute the body if not in Python 2.

jamadden:
> Wasn't sure whether to change the tests file to import PYTHON2 globally in order to only run test_clear if not running with Python 2. Made the test run always but only execute the body if not in Python 2.

Great question. I tend to be of the opinion that it's generally best to do specific feature tests rather than rely on preset broad conditions. So I would probably write that test something like this:

```python
def test_clear(self):
    inst = self._makeOne([1])
    if not hasattr(inst, 'clear'):
        raise unittest.SkipTest("PersistentList has no clear() method")
    self.assertFalse(inst._p_changed)
    inst.clear()
    self.assertEqual(inst, [])
    self.assertTrue(inst._p_changed)   
```

That makes it easier to add an implementation of `clear()` in the future on versions where it's missing without accidentally losing test coverage.

Now that I think of it, that probably wouldn't be a bad thing in and of itself (in a following PR):
```python
class PersistentList(UserList):
    __super_clear = (UserList.clear 
        if hasattr(UserList, 'clear') 
        else lambda inst: del inst[:])

    def clear(self):
         self.__super_clear()
        self._p_changed = 1
```

lungj:
The addition of a `clear()` method is actually what I wrote at first :)

persistent/list.py
```diff
@@ -37,14 +46,23 @@ class PersistentList(UserList, persistent.Persistent):
     __super_reverse = UserList.reverse
     __super_sort = UserList.sort
     __super_extend = UserList.extend
+    __super_clear = (
+        UserList.clear
+        if hasattr(UserList, 'clear')
+        else lambda inst: inst.__delitem__(slice(None, None, None))
```
tseaver: While not use the `_SLICE_ALL` defined above?

persistent/list.py
```diff
@@ -55,8 +73,19 @@ def __setslice__(self, i, j, other):
...
+            # On Python 2, the slice dunders, like ``__delslice__``,
+            # are documented as getting ``(0, sys.maxsize)`` as the
+            # arguments for list[:]. Prior to 2.7.10, it was
+            # incorrectly documented as ``(0, sys.maxint)``, but that
+            # usually worked because ``sys.maxsize`` and
+            # ``sys.maxint`` are usually the same. But on 64-bit
+            # Windown, where ``sizeof(int) == sizeof(long) == 4``,
+            # they're different (``sys.maxsize`` is bigger). See
+            # https://bugs.python.org/issue23645
+            needs_changed = i == 0 and j == sys.maxsize and bool(self)
```
tseaver: I think this change now masks "normal" list slice deetions (e.g., `del user_list[3:4]`) as being marked changed.

persistent/tests/test_list.py
```diff
@@ -268,13 +268,22 @@ def test_setslice(self):
...
         self.assertTrue(inst._p_changed)
 
-    def test_delslice(self):
+    def test_delslice_nonempty(self):
         inst = self._makeOne([1, 2, 3])
         self.assertFalse(inst._p_changed)
         self.assertEqual(inst, [1, 2, 3])
         del inst[:]
+        self.assertEqual(inst, [])
         self.assertTrue(inst._p_changed)
```
tseaver: We need a test here for a non-empty list with a non-"all" slice.

jamadden:
Thanks for the review. I will investigate tomorrow.

persistent/tests/test_list.py
jamadden: Tests and fixes added in https://github.com/zopefoundation/persistent/pull/128

PR #118, opened 2019-11-01
Enable Python 3.8 on Appveyer and also for manylinux builds

### LINKED ISSUE #117: PyPI wheels for Python 3.8
It would be nice to have Python 3.8 wheels for Windows at least (Linux users tend to have C compilers) published on PyPI.

This is a bit complicated by Appveyor not supporting Python 3.8 yet (https://github.com/appveyor/ci/issues/3142), but I expect that to happen within the next couple of days.  In any case it's possible to use 3.8 on Appveyor today if you do something like this: https://github.com/mgedmin/check-python-versions/commit/162f908c873853e446dc17feda2c1bcb904172a7

mgedmin:
Since 3.8 is not available on Appveyor yet, use a script to install it.

Meant to fix #117 eventually.


mgedmin:
That made me go and check if the manylinux docker images contain Python 3.8 already, and they do!

PR #119, opened 2019-11-06
Fix manylinux wheel builds and build them always

### LINKED ISSUE #113: Building Linux wheels seems to be broken
See https://travis-ci.org/zopefoundation/persistent/builds/530228105

mgedmin:
### Highlights

- Try manylinux builds on every commit so we can discover breakages in advance
- Fix manylinux builds so they actually work (i.e. don't upgrade wheel to latest version, which is incompatible with auditwheel)
- Add pretty names to Travis build matrix rows

Should fix #113.

### Details

The diff is messy, for which I apologize. I had to reformat the YAML to understand it.

Drop dist: xenial which has become the default.

Rename matrix: to jobs: because that's the preferred spelling according to Travis CI docs.

Move regular Linux Python builds outside build matrix, for conciseness (unfortunately this means 3.6 builds are now done twice: once without -fwrapv and once with).

Move docker pull and twine upload out of .manylinux.sh and into .travis.yml. (The docker pull mostly because we need to override the install: step and keeping it blank looks bad, and we need to override the step because we can no longer do an exit 0 in before_install because we want to do the twine upload in after_success)

Unify twine upload calls by supplying username/password via environment variables.

The changes largely mirror those in zopefoundation/BTrees#115.

### DIFF
--- a/.manylinux-install.sh
+++ b/.manylinux-install.sh
@@ -9,9 +9,7 @@ for PYBIN in /opt/python/*/bin; do
        [[ "${PYBIN}" == *"cp36"* ]] || \
        [[ "${PYBIN}" == *"cp37"* ]] || \
        [[ "${PYBIN}" == *"cp38"* ]]; then
-        # NB: it's risky to upgrade wheel without upgrading auditwheel,
-        # see https://github.com/pypa/auditwheel/issues/206
-        "${PYBIN}/pip" install -U pip setuptools wheel cffi auditwheel
+        "${PYBIN}/pip" install -U pip setuptools cffi
         "${PYBIN}/pip" install -e /io/
         "${PYBIN}/pip" wheel /io/ -w wheelhouse/
         rm -rf /io/build /io/*.egg-info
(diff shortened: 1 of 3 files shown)


mgedmin:
I'd like to restructure our travis.yml along the lines of https://github.com/zopefoundation/BTrees/pull/115 to always build wheels (and condition just the twine upload part on git tags), but that's for a separate PR.

mgedmin:
I don't want to merge this without seeing that it actually works.

I'm betting on pip being smart enough to honor the wheel version restrictions imposed by auditwheel when I specify both wheel and auditwheel on the command line, but I haven't actually verified that pip does that.

mgedmin:
Yeah, the fix wasn't working at all, because pip is much less smart than I thought it might be.

Trying an alternative fix.

mgedmin:
Ah, coveralls.  I think it's not merging the results from different Python version properly, then it sees that some branches are Python-2-only or Python-3-only and shouts.

PR #122, opened 2020-01-17
Use Python 3.6.2 on MacOS

mgedmin:
Our Python 3.6 builds on MacOS started failing with a
> Couldn't use data file '/Users/travis/build/zopefoundation/persistent/.coverage': Safety level
may not be changed inside a transaction

error from `coverage run`, after the test suite is done and it's time to collate and store coverage data.  This is probably caused by a new release of coverage.py (5.0.3; the last successfull built had coverage 4.5.4) that uses sqlite for .coverage and is reported upstream as https://github.com/nedbat/coveragepy/issues/703.

Other platforms and other Python versions do not have this problem.  Other platforms use a newer Python 3.6.x dot release, while MacOS is pinned to 3.6.0.

This PR is an experiment to see if a different Python 3.6.x dot release helps.  Turns out it does!

We have a few other zopefoundation packages using Python 3.6.1 for MacOS wheel building:
- BTrees
- zodbpickle
- zope.index

Whereas these are pinned to 3.6.0 (and I wonder if we're about to see similar failures):
- zope.container
- zope.hookable
- zope.i18nmessageid
- zope.interface
- zope.proxy
- zope.security

I'd appreciate additional attention from @jamadden and @fgregg, since IIRC you are familiar with Mac OS subtleties.  (Wasn't Python 3.6.1 the release that broke ABI compatibility with 3.6.0, and 3.6.2 again broke ABI compatibility with 3.6.1 to restore compatibility with 3.6.0?  Should we use 3.6.2 for wheel building instead of 3.6.1?)

I don't use Macs myself, but I care about Travis builds on git master being green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,9 @@ jobs:
       os: osx
       language: generic
       # NB: 3.6.0 causes https://github.com/nedbat/coveragepy/issues/703
-      env: TERRYFY_PYTHON='macpython 3.6.1'
+      # NB: 3.6.1 had that ABI regression (fixed in 3.6.2) and would be a bad
+      # version to use
+      env: TERRYFY_PYTHON='macpython 3.6.2'
     - name: Python 3.7 wheels for MacOS
       os: osx
       language: generic
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,8 @@ jobs:
     - name: Python 3.6 wheels for MacOS
       os: osx
       language: generic
-      env: TERRYFY_PYTHON='macpython 3.6.0'
+      # NB: 3.6.0 causes https://github.com/nedbat/coveragepy/issues/703
+      env: TERRYFY_PYTHON='macpython 3.6.1'
     - name: Python 3.7 wheels for MacOS
       os: osx
       language: generic


jamadden:
Yes, [3.6.1 broke the ABI](https://github.com/zopefoundation/zope.interface/pull/108) (for all platforms). I've confirmed the fix is present in 3.6.2 and later. So I wouldn't recommend pinning to 3.6.1 if we have something later available. (It's entirely possible none of our projects use the affected symbol, `PySlice_GetIndicesEx`, in which case it should be a  non-issue.)

I've been trying to track down the `sqlite3` or SQLite change that might be responsible for the coverage breakage and so far I've come up empty. 

I haven't found anything obvious in the SQLite release notes going back quite a ways. Any idea if the version of SQLite that the two Python versions were compiled against changed? 

As far as CPython goes, the diff between 3.6.0 and 3.6.1 is quite small and doesn't seem like it would be responsible, although it *does* change the exact circumstances under when new transactions are automatically begin in autocommit mode. If that's responsible for the fix, I would tend to consider this a lurking bug in coverage.py. 

```diff
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
modified: Modules/_sqlite/cursor.c
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
@@ -511,10 +511,9 @@ PyObject* _pysqlite_query_execute(pysqlite_Cursor* self, int multiple, PyObject*
     pysqlite_statement_reset(self->statement);
     pysqlite_statement_mark_dirty(self->statement);

-    /* For backwards compatibility reasons, do not start a transaction if a
-       DDL statement is encountered.  If anybody wants transactional DDL,
-       they can issue a BEGIN statement manually. */
-    if (self->connection->begin_statement && !sqlite3_stmt_readonly(self->statement->st) && !self->statement->is_ddl) {
+    /* We start a transaction implicitly before a DML statement.
+       SELECT is the only exception. See #9924. */
+    if (self->connection->begin_statement && self->statement->is_dml) {
         if (sqlite3_get_autocommit(self->connection->db)) {
             result = _pysqlite_connection_begin(self->connection);
             if (!result) {
@@ -609,7 +608,7 @@ PyObject* _pysqlite_query_execute(pysqlite_Cursor* self, int multiple, PyObject*
             }
         }

-        if (!sqlite3_stmt_readonly(self->statement->st)) {
+        if (self->statement->is_dml) {
             self->rowcount += (long)sqlite3_changes(self->connection->db);
         } else {
             self->rowcount= -1L;
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
modified: Modules/_sqlite/statement.c
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
@@ -73,8 +73,9 @@ int pysqlite_statement_create(pysqlite_Statement* self, pysqlite_Connection* con
     Py_INCREF(sql);
     self->sql = sql;

-    /* determine if the statement is a DDL statement */
-    self->is_ddl = 0;
+    /* Determine if the statement is a DML statement.
+       SELECT is the only exception. See #9924. */
+    self->is_dml = 0;
     for (p = sql_cstr; *p != 0; p++) {
         switch (*p) {
             case ' ':
@@ -84,9 +85,10 @@ int pysqlite_statement_create(pysqlite_Statement* self, pysqlite_Connection* con
                 continue;
         }

-        self->is_ddl = (PyOS_strnicmp(p, "create ", 7) == 0)
-                    || (PyOS_strnicmp(p, "drop ", 5) == 0)
-                    || (PyOS_strnicmp(p, "reindex ", 8) == 0);
+        self->is_dml = (PyOS_strnicmp(p, "insert ", 7) == 0)
+                    || (PyOS_strnicmp(p, "update ", 7) == 0)
+                    || (PyOS_strnicmp(p, "delete ", 7) == 0)
+                    || (PyOS_strnicmp(p, "replace ", 8) == 0);
         break;
     }
```

fgregg:
let's pin it to 3.6.2 as @jamadden.

mgedmin:
> (It's entirely possible none of our projects use the affected symbol, PySlice_GetIndicesEx, in which case it should be a non-issue.

BTrees/BTrees/BTreeItemsTemplate.c calls PySlice_GetIndicesEx.  BTrees is also one of the projects that are building Mac OS wheels on Python 3.6.1.

ripgrep finds no other calls to PySlice_GetIndicesEx in my ~/src/zopefoundation/*.

mgedmin:
Pinned to 3.6.2.  CI passes.

PR #123, opened 2020-02-12
Try out Python 3.9 on TravisCI.

icemac:


icemac:
Before we can declare Python 3.9 support #124 has to be fixed.

mgedmin:
Also, there's appveyor.yml and manylinux scripts.

icemac:
I added 3.9 to the manylinux script but Appveyor does not yet have Python 3.9.

I'd like to merge this PR before making a release to hopefully get a step further towards Python 3.9 compatibility in Zope and ZTK.

mgedmin:
> I added 3.9 to the manylinux script but Appveyor does not yet have Python 3.9.

Our appveyor.yml uses https://raw.githubusercontent.com/matthew-brett/multibuild/11a389d78892cf90addac8f69433d5e22bfa422a/install_python.ps1 to download and install missing Pythons from www.python.org.  That script supports release candidates (PYTHON being set to "c:\python39rc1" etc), but it currently doesn't handle alpha/beta releases.

We could fork that script and add support for Pythons like c:\python39a4, if we want to.

icemac:
@mgedmin You convinced me: Lets only try Python 3.9 support on Travis, but do not create wheels and do not declare we support it until a final release.

PR #128, opened 2020-02-15
Fix slicing and copying of PersistentList and PersistentMapping.

### LINKED ISSUE #112: PersistentList slicing returns a non-persistent list on python3
Since `__setslice__()` and the likes have been killed in python3 (in favor of `__getitem__(slice(...)`), it seems that PersistentList must be updated, else slices return regular lists and the DB gets dangerously filled with non-persistent mutables:

```pycon
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from persistent.list import PersistentList
>>> a = PersistentList([1, 2,3])
>>> isinstance(a, PersistentList)
True
>>> isinstance(a[1:], PersistentList)
False
```

jamadden:
See https://bugs.python.org/issue27639 and https://bugs.python.org/issue27141

Fixes #112

### DIFF
--- a/persistent/list.py
+++ b/persistent/list.py
@@ -31,8 +31,8 @@ class PersistentList(UserList, persistent.Persistent):
     as changed and automatically persisted.
 
     .. versionchanged:: 4.5.2
-       Using the `clear` method, or writing ``del inst[:]`` now only
-       results in marking the instance as changed if it actually removed
+       Using the `clear` method, or deleting a slice (e.g., ``del inst[:]`` or ``del inst[x:x]``)
+       now only results in marking the instance as changed if it actually removed
        items.
     .. versionchanged:: 4.5.2
        The `copy` method is available on Python 2.
@@ -52,7 +52,7 @@ class PersistentList(UserList, persistent.Persistent):
     __super_clear = (
         UserList.clear
         if hasattr(UserList, 'clear')
-        else lambda inst: inst.__delitem__(slice(None, None, None))
+        else lambda inst: inst.__delitem__(_SLICE_ALL)
     )
 
     if (2, 7, 0) < sys.version_info[:3] < (3, 7, 4):
@@ -84,10 +84,11 @@ class PersistentList(UserList, persistent.Persistent):
 
     def __delitem__(self, i):
         # If they write del list[:] but we're empty,
-        # no need to mark us changed.
-        needs_changed = i != _SLICE_ALL or bool(self)
+        # no need to mark us changed. Likewise with
+        # a slice that's empty, like list[1:1].
+        len_before = len(self.data)
         self.__super_delitem(i)
-        if needs_changed:
+        if len(self.data) != len_before:
             self._p_changed = 1
 
     if PYTHON2:  # pragma: no cover
@@ -102,19 +103,11 @@ class PersistentList(UserList, persistent.Persistent):
             self._p_changed = 1
 
         def __delslice__(self, i, j):
-            # On Python 2, the slice dunders, like ``__delslice__``,
-            # are documented as getting ``(0, sys.maxsize)`` as the
-            # arguments for list[:]. Prior to 2.7.10, it was
-            # incorrectly documented as ``(0, sys.maxint)``, but that
-            # usually worked because ``sys.maxsize`` and
-            # ``sys.maxint`` are usually the same. But on 64-bit
-            # Windown, where ``sizeof(int) == sizeof(long) == 4``,
-            # they're different (``sys.maxsize`` is bigger). See
-            # https://bugs.python.org/issue23645
-            needs_changed = i == 0 and j == sys.maxsize and bool(self)
-            self.__super_delslice(i, j)
-            if needs_changed:
-                self._p_changed = 1
+            # In the past we just called super, but we want to apply the
+            # same _p_changed optimization logic that __delitem__ does. Don't
+            # call it as ``self.__delitem__``, though, because user code in subclasses
+            # on Python 2 may not be expecting to get a slice.
+            PersistentList.__delitem__(self, slice(i, j))
 
     def __iadd__(self, other):
         L = self.__super_iadd(other)
--- a/persistent/list.py
+++ b/persistent/list.py
@@ -55,8 +55,8 @@ class PersistentList(UserList, persistent.Persistent):
         else lambda inst: inst.__delitem__(_SLICE_ALL)
     )
 
-    if (2, 7, 0) < sys.version_info[:3] < (3, 7, 4):
-        # Prior to 3.7.4, Python 3 failed to properly
+    if not PYTHON2 and sys.version_info[:3] < (3, 7, 4):
+        # Prior to 3.7.4, Python 3 (but not Python 2) failed to properly
         # return an instance of the same class.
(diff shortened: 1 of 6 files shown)


persistent/list.py
```diff
@@ -49,43 +52,62 @@ class PersistentList(UserList, persistent.Persistent):
     __super_clear = (
         UserList.clear
         if hasattr(UserList, 'clear')
-        else lambda inst: inst.__delitem__(slice(None, None, None))
+        else lambda inst: inst.__delitem__(_SLICE_ALL)
     )
 
+    if (2, 7, 0) < sys.version_info[:3] < (3, 7, 4):
```
tseaver: Don't think we care about 2.6.x. Simpler as:
```suggestion
    if sys.version_info[:3] < (3, 7, 4):
```


or did you mean to be leaving the native `__getitem__` in place for 2.7.x?  In which case:
```suggestion
    if (3, 0. 0) < sys.version_info[:3] < (3, 7, 4):
```

persistent/list.py
```diff
@@ -49,43 +52,62 @@ class PersistentList(UserList, persistent.Persistent):
...
     def __delitem__(self, i):
         # If they write del list[:] but we're empty,
-        # no need to mark us changed.
-        needs_changed = i != _SLICE_ALL or bool(self)
+        # no need to mark us changed. Likewise with
+        # a slice that's empty, like list[1:1].
+        len_before = len(self.data)
         self.__super_delitem(i)
-        if needs_changed:
+        if len(self.data) != len_before:
```
tseaver: :+1:

persistent/list.py
```diff
@@ -49,43 +52,62 @@ class PersistentList(UserList, persistent.Persistent):
...
             self._p_changed = 1
 
     if PYTHON2:  # pragma: no cover
         __super_setslice = UserList.__setslice__
         __super_delslice = UserList.__delslice__
 
+        def copy(self):
+            return self.__class__(self)
```
tseaver: :+1:

persistent/list.py
```diff
@@ -49,43 +52,62 @@ class PersistentList(UserList, persistent.Persistent):
...
-            # https://bugs.python.org/issue23645
-            needs_changed = i == 0 and j == sys.maxsize and bool(self)
-            self.__super_delslice(i, j)
-            if needs_changed:
-                self._p_changed = 1
+            # In the past we just called super, but we want to apply the
+            # same _p_changed optimization logic that __delitem__ does. Don't
+            # call it as ``self.__delitem__``, though, because user code in subclasses
+            # on Python 2 may not be expecting to get a slice.
+            PersistentList.__delitem__(self, slice(i, j))
```
tseaver: :+1:

persistent/mapping.py
```diff
@@ -55,6 +57,17 @@ class PersistentMapping(IterableUserDict, persistent.Persistent):
...
 
+    if sys.version_info[:3] < (3, 7, 4):
+        # See PersistentList.
+        # See https://github.com/python/cpython/commit/3645d29a1dc2102fdb0f5f0c0129ff2295bcd768
+        def __copy__(self):
+            inst = self.__class__.__new__(self.__class__)
+            inst.__dict__.update(self.__dict__)
+            # Create a copy and avoid triggering descriptors
+            data = self.__dict__['data'] if 'data' in self.__dict__ else self.__dict__['_containers']
+            inst.__dict__["data"] = data.copy()
```
tseaver: I'm not sure why the `inst.update` call isn't sufficient above. This bit seems to make *two* copies of the underlying data, one under `data` and one under `_containers`.   Actually, where does `_containers` come from?

```python
Python 2.7.17 (default, Feb 11 2020, 17:51:09) 
[GCC 7.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from UserDict import UserDict
>>> class Foo(UserDict): pass
... 
>>> foo = Foo()
>>> foo.__dict__
{'data': {}}
```

```python
$ python3.6
Python 3.6.10 (default, Feb 11 2020, 17:35:51) 
[GCC 7.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import UserDict
>>> 
>>> class Foo(UserDict): pass
... 
>>> foo = Foo()
>>> foo.__dict__
{'data': {}}
>>> foo['bar'] = 0
>>> foo.__dict__
{'data': {'bar': 0}}
```

```python
Python 3.7.6 (default, Feb 11 2020, 17:30:33) 
[GCC 7.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import UserDict
>>> class Foo(UserDict): pass
... 
>>> foo = Foo()
>>> foo.__dict__
{'data': {}}
>>> foo['bar'] = 0
>>> foo.__dict__
{'data': {'bar': 0}}
```

persistent/list.py
jamadden: Ah, thank you. I did mean to leave the original version in place for Python 2. This condition was left over from an earlier iteration (it was originally combined with the condition below).

persistent/mapping.py
jamadden: > I'm not sure why the inst.update call isn't sufficient above.

Because `update` makes a shallow copy, and the second object still shares `data` with the first, meaning they both mutate when you change only one. This is basically a backport of the fix for https://bugs.python.org/issue27141. (It's less of a big deal with persistent objects because you're mutating one to probably persist (just) it again and the copy will be made at persistence time, but still a gotcha.)

> one under data and one under _containers

You're right, I should probably pop `_container` from the copied dict.

> Actually, where does _containers come from?

The mists of history; it's been present since the first revision in this repository. 

https://github.com/zopefoundation/persistent/blob/a6a18eaa8f347530646ce413f93f90a584ab2fb2/persistent/mapping.py#L124-L134

It must be something specific to `persistent`, though, because it doesn't show up in the [very first revision of `UserDict` from CPython](https://github.com/python/cpython/blame/bd40d7e69ffd53f36a315605fc849bb12ae1f55f/Lib/UserDict.py)

jamadden: It goes back to [this change from 2001](https://github.com/zopefoundation/ZODB/commit/c64281cf2830b569eed4f211630a8a61d22a0f0b#diff-b0f568e20f51129c10a096abad27c64a). Apparently "Standalone ZODB" used `data` but `ZODB` proper used `_container`. 

I wonder if there are any pickles that have been read-but-not-written (and are still being used read-only) with all the versions of ZODB and persistent since 2001. Surely not. Maybe this whole thing can be dropped.

jamadden:
Rebased on master and comments addressed.

persistent/mapping.py
mgedmin: > Surely not

Hahahahaha.

I admire your optimism.

jamadden: Hope springs eternal. (I mean, there could be people entering university that are younger than those pickles!)

I added specific tests for the copy functionality when the ultra-legacy data layout is in place. I *hope* this PR is getting close to approval 😄

jamadden:
Thank you!

PR #129, opened 2020-02-18
Fix issue 124

### LINKED ISSUE #124: cPickleCache will not build on Python 3.9 without trace refs
cPickleCache is relying on https://github.com/zopefoundation/persistent/blob/d4f1204d87bec113de5c2457eaa7878ec0f40ab3/persistent/cPickleCache.c#L664

which is no longer available on Python 3.9 without Py_TRACE_REFS being defined, s. https://github.com/python/cpython/pull/18332

As a result, cPickleCache can't be built + imported:

```
from persistent import cPickleCache
*** ImportError: /home/thomas/devel/plone/persistent/.tox/py39/lib/python3.9/site-packages/persistent/cPickleCache.cpython-39-x86_64-linux-gnu.so: undefined symbol: _Py_ForgetReference
```

Tests silently work around this.

It is still possible to import cPersistence though, so code in the package that does `import persistence` and then uses `persistent.Persistent`, e.g. https://github.com/zopefoundation/persistent/blob/d4f1204d87bec113de5c2457eaa7878ec0f40ab3/persistent/mapping.py#L33 will be fine. But client code that uses the items defined in __init__.py:

https://github.com/zopefoundation/persistent/blob/d4f1204d87bec113de5c2457eaa7878ec0f40ab3/persistent/__init__.py#L40

will only ever see the Python implementation of both Persistent and the pickle cache.

-> This is causing trouble in the Persistence package, ticket coming there.

jamadden:
There are three commits showing the iterations I went through.

- The first commit adds the necessary preprocessor magic to make the extension build and run on 3.9.0a3+.
- The second commit adds more preprocessor magic to make it build and run on 3.9.0a3 if `Py_TRACE_REFS` is actually defined.
- The third commit removes all usage of those internal functions entirely, after I verified that everything still worked on the `Py_TRACE_REFS` build.

Rationale to remove usage of the internal functions:

- They're documented specifically as only to be used by the core interpreter.
- The nesting of conditions to make them work was pretty ugly.
- Especially since the core logic of the function is already "hairy."
- It was not at all clear to me that they were actually doing the right thing anymore given the way the interpreter has evolved (e.g., with the separation of `Py_REF_DEBUG` and `Py_TRACE_REFS`); it's possible they could do more harm than good (I didn't observe any issues with just `Py_TRACE_REFS`, but I didn't test any more combos of settings)
- I *think* they were there to assist in debugging. But this function hasn't changed substantially in the history of this repository (9 years), so maybe it's debugged.

Fixes #124

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -290,7 +290,7 @@ changed(cPersistentObject *self)
         }
         Py_INCREF(self);
         PyTuple_SET_ITEM(arg, 0, (PyObject *)self);
-        result = PyEval_CallObject(meth, arg);
+        result = PyObject_CallObject(meth, arg);
         Py_DECREF(arg);
         Py_DECREF(meth);
         if (result == NULL)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,8 +19,8 @@
   ``PersistentMapping.popitem`` to no longer mark the object as
   changed if it was empty.
 
-- Add support for Python 3.9a3+ by updating references to deprecated
-  and moved functions. See `issue 124 <https://github.com/zopefoundation/persistent/issues/124>`_.
+- Add support for Python 3.9a3+.
+  See `issue 124 <https://github.com/zopefoundation/persistent/issues/124>`_.
 
 
 4.5.1 (2019-11-06)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,10 @@
   ``PersistentMapping.popitem`` to no longer mark the object as
   changed if it was empty.
 
+- Add support for Python 3.9a3+ by updating references to deprecated
+  and moved functions. See `issue 124 <https://github.com/zopefoundation/persistent/issues/124>`_.
+
+
 4.5.1 (2019-11-06)
 ------------------
 
--- a/persistent/cPickleCache.c
+++ b/persistent/cPickleCache.c
@@ -627,30 +627,7 @@ cc_oid_unreferenced(ccobject *self, PyObject *oid)
         to decref an already deleted object.
     */
 
-#if defined(Py_TRACE_REFS) && !defined(Py_LIMITED_API)
-    /* This is called from the deallocation function after the
-        interpreter has untracked the reference.  Track it again.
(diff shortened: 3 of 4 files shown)


persistent/cPersistence.c
```diff
@@ -290,7 +290,7 @@ changed(cPersistentObject *self)
         }
         Py_INCREF(self);
         PyTuple_SET_ITEM(arg, 0, (PyObject *)self);
-        result = PyEval_CallObject(meth, arg);
+        result = PyObject_CallObject(meth, arg);
```
jamadden: `PyEval_CallObject` is deprecated in 3.9. `PyObject_CallObject` is a drop-in replacement AFAICS. (An earlier iteration of the change note mentioned something about deprecated functions but I dropped that.)

icemac:
I like it very much that you tackled this issue although I am not able to review it.
Maybe it should be tested on Python 3.9 on TravisCI as well.

icemac:
There is #123 where I ran the tests on Python 3.9 on Travis CI but they where green without the changes here. According to #124 this is because the pure Python version us used. Can we do something to force using the C version or at least check if it can be used to have a more clear indicator for changes in the C API in future?

jamadden:
> I ran the tests on Python 3.9 on Travis CI but they where green without the changes here. According to #124 this is because the pure Python version us used

In this case, it's because the breaking changes are not actually released yet. They are not in 3.9a3, only in git master. Everything compiles and runs fine (except for that one warning) on 3.9a3 still.

> Can we do something to force using the C version 

I think so. I like the way `zope.interface` is handling things (a decorator, plus a distinction between `PURE_PYTHON=0` and `PURE_PYTHON=1`) and I'd like to spread that around.

jamadden:
> have a more clear indicator for changes in the C API in future?

We could also try setting `CFLAGS=-Werror`. That would have caught the `PyEval_CallObject` issue, but, since `_Py_NewReference` and friends are private, it wouldn't have caught that. Plus there's the possibility of false positives unless it's scoped very carefully.

jamadden:
Because this is such a simple change that only affects specialized builds of Python, I'm tempted to merge it.

PR #132, opened 2020-02-20
Add the ability to force use of C extensions.

### LINKED ISSUE #131: Support ability to require use of C extensions
This has been a problem for end users and our own testing (cf #124 and https://github.com/zopefoundation/persistent/pull/129#issuecomment-588081525).

I propose doing basically what [zope.interface now does](https://github.com/zopefoundation/zope.interface/pull/151/)

jamadden:
With PURE_PYTHON=0, like in zope.interface.

Also always require all three extensions. This solves mysterious issues you can get if you wind up mixing and matching (#124).

Fixes #131

Add travis and tox tests for this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ env:
     - TWINE_USERNAME: zope.wheelbuilder
     # this sets $PYPIPASSWORD
     - secure: "NTWzDr5p8KRPNt+sniTot7csbzC87rzir/XfLtENE0GpQ49FlKw3lBhsDqAPoD8Ea5lwiHXmC/C/ci1UZhFvVEkAoQ2qJlMRnhqUdRJSrqcitmRt0fT6mLaTd+Lr+DxKlBxpssobrEm2G42V/G1s0Ggym04OqF8T+s6MF5ywgJM="
+    # We want to require the C extensions to build and function
+    # everywhere (except where we specifically opt-out, currently just
+    # PyPy, where they build but don't quite work).
+    - PURE_PYTHON: 0
+
 
 python:
   - 2.7
@@ -12,11 +17,15 @@ python:
   - 3.6
   - 3.7
   - 3.8
-  - pypy
-  - pypy3
 
 jobs:
   include:
+    # Don't test C extensions on PyPy.
+    - python: pypy
+      env: PURE_PYTHON=1
+
+    - python: pypy3
+      env: PURE_PYTHON=1
 
     # Special Linux builds
     - name: "Python: 2.7, pure (no C extensions)"
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,8 +12,8 @@
   broken prior to Python 3.7.4.
 
 - Update the handling of the ``PURE_PYTHON`` environment variable.
-  Now, a value of 0 requires the C extensions to be used; any other
-  value prevents the extensions from being used. Also, all C
+  Now, a value of "0" requires that the C extensions be used; any other
+  non-empty value prevents the extensions from being used. Also, all C
   extensions are required together or none of them will be used. This
   prevents strange errors that arise from a mismatch of Python and C
   implementations. See `issue 131 <https://github.com/zopefoundation/persistent/issues/131>`_.
@@ -22,7 +22,8 @@
   the pure-Python implementations have changed.
 
 - Fix ``PersistentList`` to mark itself as changed after calling
-  ``clear``. See `PR 115 <https://github.com/zopefoundation/persistent/pull/115/>`_.
+  ``clear`` (if needed). See `PR 115
+  <https://github.com/zopefoundation/persistent/pull/115/>`_.
 
 - Fix ``PersistentMapping.update`` to accept keyword arguments like
   the native ``UserDict``. Previously, most uses of keyword arguments
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 ``persistent`` Changelog
 ========================
 
-4.5.2 (unreleased)
+4.6.0 (unreleased)
 ------------------
 
 - Fix slicing of ``PersistentList`` to always return instances of the
@@ -10,6 +10,16 @@
 - Fix copying  of ``PersistentList`` and ``PersistentMapping`` using
   ``copy.copy`` to also copy the underlying data object. This was
   broken prior to Python 3.7.4.
+
+- Update the handling of the ``PURE_PYTHON`` environment variable.
+  Now, a value of 0 requires the C extensions to be used; any other
+  value prevents the extensions from being used. Also, all C
+  extensions are required together or none of them will be used. This
+  prevents strange errors that arise from a mismatch of Python and C
+  implementations. See `issue 131 <https://github.com/zopefoundation/persistent/issues/131>`_.
+
+  Note that some private implementation details such as the names of
+  the pure-Python implementations have changed.
 
 - Fix ``PersistentList`` to mark itself as changed after calling
   ``clear``. See `PR 115 <https://github.com/zopefoundation/persistent/pull/115/>`_.
--- a/persistent/_compat.py
+++ b/persistent/_compat.py
@@ -56,7 +56,7 @@ def _c_optimizations_required():
     """
     Return a true value if the C optimizations are required.
 
-    This uses the ``PURE_PYTHON`` variable as documented in `_use_c_impl`.
+    This uses the ``PURE_PYTHON`` variable as documented in `use_c_impl`.
     """
     pure_env = os.environ.get('PURE_PYTHON')
     require_c = pure_env == "0"
@@ -91,10 +91,17 @@ def _c_optimizations_available():
 def _c_optimizations_ignored():
     """
     The opposite of `_c_optimizations_required`.
+
+    On PyPy, this always returns True.
+
+    Otherwise, if ``$PURE_PYTHON`` is set to any non-empty value
+    besides "0", optimizations are ignored. Setting ``$PURE_PYTHON``
+    to "1", for example, ignores optimizations. Setting ``$PURE_PYTHON`` to
+    an empty value *does not* ignore optimizations.
     """
     pure_env = os.environ.get('PURE_PYTHON')
     # The extensions can be compiled with PyPy 7.3, but they don't work.
-    return PYPY or (pure_env is not None and pure_env != "0")
+    return PYPY or (pure_env and pure_env != "0")
 
 
 def _should_attempt_c_optimizations():
@@ -102,7 +109,10 @@ def _should_attempt_c_optimizations():
     Return a true value if we should attempt to use the C optimizations.
 
     This takes into account whether we're on PyPy and the value of the
-    ``PURE_PYTHON`` environment variable, as defined in `_use_c_impl`.
+    ``PURE_PYTHON`` environment variable, as defined in `use_c_impl`.
+
+    Note that setting ``PURE_PYTHON=0`` forces the use of C optimizations,
+    even on PyPy.
     """
     if _c_optimizations_required():
         return True
@@ -111,7 +121,7 @@ def _should_attempt_c_optimizations():
     return not _c_optimizations_ignored()
 
 
-def use_c_impl(py_impl, name=None, globs=None):
+def use_c_impl(py_impl, name=None, globs=None, mod_name=None):
     """
     Decorator. Given an object implemented in Python, with a name like
     ``Foo``, import the corresponding C implementation from
@@ -163,7 +173,7 @@ def use_c_impl(py_impl, name=None, globs=None):
     """
     name = name or py_impl.__name__
     globs = globs or sys._getframe(1).f_globals
-    mod_name = globs['__name__']
+    mod_name = mod_name or globs['__name__']
 
     def find_impl():
         if not _should_attempt_c_optimizations():
--- a/persistent/_compat.py
+++ b/persistent/_compat.py
@@ -14,9 +14,23 @@
 
 import sys
 import os
+import types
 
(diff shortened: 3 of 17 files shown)


persistent/cPersistence.c
```diff
@@ -1270,6 +1271,18 @@ Per_get_mtime(cPersistentObject *self)
         return Py_None;
     }
 
+    if (!TimeStamp)
+    {
+        PyObject* ts_module;
+        ts_module = PyImport_ImportModule("persistent._timestamp");
```
jamadden: This had to move from module initialization time because it created a cycle.

Note that it's importing the C version only now. 

Also note the new error handling at the end (`if !TimeStamp`). That turned out to be a problem that resulted in a `SystemError` at one point; that shouldn't be possible now that we're getting the C implementation, however.

persistent/persistence.py
```diff
@@ -52,62 +54,69 @@
...
 _OID_STRUCT = struct.Struct('>Q')
 _OID_UNPACK = _OID_STRUCT.unpack
 
-@implementer(IPersistent)
-class Persistent(object):
+# We don't apply ``use_c_impl`` as a decorator to a class named
+# Persistent because we need access to the PersistentPy name
+# in a few places in the implementation.
+@implementer(interfaces.IPersistent)
+class PersistentPy(object):
```
jamadden: I went back and forth on this. I originally stuck with the name `Persistent`, but it looked weird (and of course caused linter errors) to refer to `PersistentPy` in the body; and if you forgot, you'd get weird errors. I only went with `PersistentPy` when I started to write a comment explaining that the class was named `Persistent` but you had to refer to `PersistentPy` in the body.

The other alternative would be to simply not import the C extension into this module at all, just in `__init__.py`. But I like the symmetry of doing everything (nearly) the same across all the modules and treating the C extension as an implementation detail private to the module.

.travis.yml
```diff
@@ -32,6 +32,10 @@ jobs:
       python: 3.7
       env: PURE_PYTHON=1
 
+    - name: "Python: 3.8, (forced C extensions)"
```
mgedmin: comma parenthesis looks weird.

In previous jobs the parentheses were explaining the meaning of "pure".  I suggest no parentheses here: "Python: 3.8, forced C extensions"

.travis.yml
```diff
@@ -32,6 +32,10 @@ jobs:
...
+      python: 3.7
```
mgedmin: The job name says 3.8!

tox.ini
```diff
@@ -32,6 +26,8 @@ basepython =
     python3.6
 commands =
     coverage run -m zope.testrunner --test-path=.
+    python -c 'import os, subprocess; env = os.environ.copy(); env["PURE_PYTHON"] = "1"; subprocess.check_call("coverage run -a -m zope.testrunner --test-path=.", shell=True, env=env)'
+    python -c 'import os, subprocess; env = os.environ.copy(); env["PURE_PYTHON"] = "0"; subprocess.check_call("coverage run -a -m zope.testrunner --test-path=.", shell=True, env=env)'
```
mgedmin: Uhh, this is for cross-platformness of some kind?

Because `env PURE_PYTHON=1 coverage run -a -m ...` would be shorter and more readable on any OS that has /usr/bin/env.  Google tells me it exist on MacOS, as well as on Linux.  Do we care about people using `tox` on Windows?

Suggestion: `import os, subprocess; subprocess.check_call(..., env=dict(os.environ, PURE_PYTHON="0"))`  to make the oneliners shorter.

.travis.yml
jamadden: It occurs to me that on all the CPythons on CI we really want to force that the C extensions are built and working; they're not optional on CI. So I need to make that the default and only opt-out on PyPy.

tox.ini
jamadden: > Uhh, this is for cross-platformness of some kind?
Right.

> Because env PURE_PYTHON=1 coverage run -a -m ... would be shorter and more readable on any OS that has /usr/bin/env. Google tells me it exist on MacOS, as well as on Linux. Do we care about people using tox on Windows?

Using `env` didn't occur to me. Doing so would require whitelisting the command to avoid the warning from tox about using an executable outside the environment, though.

> Suggestion: import os, subprocess; subprocess.check_call(..., env=dict(os.environ, PURE_PYTHON="0")) to make the oneliners shorter.

Oh, I like that. Thank you. I don't think I've ever actually used (or even come across!) that `dict` constructor.

jamadden:
Rebased on master.

icemac:
TravisCI is green now (after restarting some jobs). Is there stil something that needs to be done?

jamadden:
Awaiting review.

mgedmin:
I still don't have the energy for reviewing this, sorry!

icemac:
Same here, sorry, too.

jamadden:
I found a way to make the diff much smaller and simpler by avoiding the need to rename `Persistent` to `PersistentPy`. (`use_c_impl` updates the globals of the Python functions so that the name the class is defined by is still the name the class bodies see when they execute.) Hopefully that helps.

persistent/timestamp.py
```diff
@@ -79,8 +79,10 @@ def _parseRaw(octets):
     second = b * _TS_SECOND_BYTES_BIAS
     return (year, month, day, hour, minute, second)
 
+TimeStampPy = None
```
mgedmin: This could use a comment explaining that it'll be overwritten by the `@use_c_impl` decorator and is here solely to prevent linter warnings, or whatever it's actually here for.

persistent/tests/test_timestamp.py
```diff
@@ -47,10 +48,10 @@ def test_fromutc(self):
         self.assertTrue(utc.fromutc(source) is source)
 
 
-class pyTimeStampTests(unittest.TestCase):
+class TimeStampTests(unittest.TestCase):
 
     def _getTargetClass(self):
-        from persistent.timestamp import pyTimeStamp
+        from persistent.timestamp import TimeStampPy as pyTimeStamp
         return pyTimeStamp
```
mgedmin: I'd change the next line to `return TimeStampPy` instead of using `import ... as` here.

persistent/tests/test_timestamp.py
```diff
@@ -198,14 +199,74 @@ def check_py3(op, passes):
                 check(op, passes)
 
 
-class TimeStampTests(pyTimeStampTests):
+class pyTimeStampTests(TimeStampTests):
+    # Things specific to the Python implementation
```
mgedmin: It would be nice to verify that the C code still returns the same hash values for the same timestamps.  (Granted, the old code didn't do that.)

The monkey-patching parts to force a 32- or 64-bit algorithm from the Python implementation is, of course, specific to the Python implementation.

persistent/tests/test_timestamp.py
```diff
@@ -225,11 +286,11 @@ def _make_many_instants(self):
...
     def _makeC(self, *args, **kwargs):
-        from persistent.timestamp import TimeStamp
-        return TimeStamp(*args, **kwargs)
+        from persistent._compat import _c_optimizations_available as get_c
+        return get_c()['persistent.timestamp'].TimeStamp(*args, **kwargs)
 
     def _makePy(self, *args, **kwargs):
-        from persistent.timestamp import pyTimeStamp
+        from persistent.timestamp import TimeStampPy as pyTimeStamp
         return pyTimeStamp(*args, **kwargs)
```
mgedmin: I would also use `return TimeStampPy(...)` here instead of `import ... as`.

persistent/tests/test_picklecache.py
```diff
@@ -1096,33 +1075,45 @@ def _p_deactivate(self):
...
 
         def makePersistent(oid):
-           persist = self._getDummyPersistentClass()()
-           persist._p_oid = oid
-           persist._p_jar = dummy_connection
-           return persist
+            persist = self._getDummyPersistentClass()()
+            persist._p_oid = oid
+            persist._p_jar = dummy_connection
+            return persist
```
mgedmin: ♥

persistent/tests/test_picklecache.py
```diff
@@ -1096,33 +1075,45 @@ def _p_deactivate(self):
...
+        from persistent.interfaces import IExtendedPickleCache
+        from zope.interface.verify import verifyObject
+        from zope.interface.exceptions import DoesNotImplement
+        from zope.interface.exceptions import BrokenImplementation
+        # We don't claim to implement it.
+        with self.assertRaises(DoesNotImplement):
+            verifyObject(IExtendedPickleCache, self._makeOne())
+        # And we don't even provide everything it asks for.
+        with self.assertRaises(BrokenImplementation):
+            verifyObject(IExtendedPickleCache, self._makeOne(), tentative=True)
```
mgedmin: That seems like a strange thing to test for.

persistent/picklecache.py
```diff
@@ -51,11 +56,21 @@ def locked(self, *args, **kwargs):
             self._is_sweeping_ring = False
     return locked
 
-from .ring import Ring
+# This name will be replaced by the use_c_impl decorator.
+PickleCachePy = None
```
mgedmin: :100: for the comment

persistent/persistence.py
```diff
@@ -11,22 +11,30 @@
...
+# we have lots of not-quite-correct continuation indentation.
+# TODO: Fix that in a whitespace-only commit.
+# pylint:disable=bad-continuation
+# There are a few places we need to work with exact types.
+# pylint:disable=unidiomatic-typecheck
 
+GHOST = use_c_impl(interfaces.GHOST, 'GHOST')
+UPTODATE = use_c_impl(interfaces.UPTODATE, 'UPTODATE')
+CHANGED = use_c_impl(interfaces.CHANGED, 'CHANGED')
+STICKY = use_c_impl(interfaces.STICKY, 'STICKY')
```
mgedmin: This looks like magic that could use an explanatory comment?

persistent/persistence.py
```diff
@@ -595,7 +613,8 @@ def __repr__(self):
         return '<%s.%s object at 0x%x%s%s%s>' % (
             # Match the C name for this exact class
             type(self).__module__ if type(self) is not Persistent else 'persistent',
-            type(self).__name__, id(self),
+            type(self).__name__ if type(self) is not Persistent else 'Persistent',
+            id(self),
```
mgedmin: Nice!

persistent/_compat.py
```diff
@@ -35,3 +48,179 @@
...
+        return {}
+
+
+def _c_optimizations_ignored():
+    """
+    The opposite of `_c_optimizations_required`.
+    """
+    pure_env = os.environ.get('PURE_PYTHON')
+    # The extensions can be compiled with PyPy 7.3, but they don't work.
+    return PYPY or (pure_env is not None and pure_env != "0")
```
mgedmin: So setting PURE_PYTHON to an empty string is equivalent to setting it to 1?  I'm not sure how I feel about that, but okay.

persistent/_compat.py
```diff
@@ -35,3 +48,179 @@
...
+        return getattr(c_opt, name)
+
+    c_impl = find_impl()
+    # Always make available by the FooPy name
+    globs[name + 'Py'] = py_impl
+
+    if c_impl is not py_impl and isinstance(py_impl, type):
+        # Rebind the globals of all the functions to still see the
+        # object under its original class name, so that references
+        # in function bodies work as expected.
```
mgedmin: Wow, clever!

persistent/tests/test_picklecache.py
mgedmin: Ah, I think I get it: it's to make sure the `@use_c_impl` doesn't accidentally copy the IExtendedPickleCache declaration onto the C version of the pickle cache that doesn't actually implement it!

persistent/persistence.py
mgedmin: Also, the Python versions are already imports from the C versions, when those are available?

persistent/_compat.py
jamadden: I was trying to preserve compatibility. Putting PURE_PYTHON in the environment *period* would result in no C extensions; the value was immaterial. 

Except as I write that, I realize that no, this package actually handled things a bit differently than, say, [zope.proxy](https://github.com/zopefoundation/zope.proxy/blob/5773b8b33e42d28962684f7647f4dca089518259/src/zope/proxy/__init__.py#L515). Instead of checking for its presence *in* the environment, it checks its value. 

https://github.com/zopefoundation/persistent/blob/29d207e7849bb55b371e1a4c6674eb4264ab9fbd/persistent/_compat.py#L18

https://github.com/zopefoundation/persistent/blob/29d207e7849bb55b371e1a4c6674eb4264ab9fbd/persistent/__init__.py#L55-L56

So maybe that can/should be simplified/corrected.

CHANGES.rst
```diff
@@ -11,6 +11,16 @@
...
 
+- Update the handling of the ``PURE_PYTHON`` environment variable.
+  Now, a value of 0 requires the C extensions to be used; any other
+  value prevents the extensions from being used. Also, all C
+  extensions are required together or none of them will be used. This
+  prevents strange errors that arise from a mismatch of Python and C
+  implementations. See `issue 131 <https://github.com/zopefoundation/persistent/issues/131>`_.
+
+  Note that some private implementation details such as the names of
+  the pure-Python implementations have changed.
```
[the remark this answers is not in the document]
jamadden: Yes. `pyTimeStamp` -> `TimeStampPy`.

persistent/tests/test_picklecache.py
jamadden: Exactly. Seems like it could use an explanatory comment.

PR #134, opened 2020-03-04
Use ffi.gc() on the ring nodes to avoid needing a weakref to PersistentPy

### LINKED ISSUE #133: Python/C inconsistency: Cannot store raw `Persistent` objects in cache
The C implementation lets you store basic `Persistent` instances in the cache:
```pycon
>>> from persistent import Persistent
>>> from persistent import PickleCache
>>> obj = Persistent()
>>> type(obj)
<class 'persistent.Persistent'>
>>> key = obj._p_oid = b'00000000'
>>> obj._p_jar = object()
>>> cache[key] = obj
>>> cache[key]
<persistent.Persistent object at 0x1053ae960 oid 0x3030303030303030 in <object object at 0x1053a8430>>
```

But the pure-Python implementation does not:

```pycon
>>> from persistent.picklecache import PickleCachePy
>>> from persistent import PersistentPy
>>> pyobj = PersistentPy()
>>> pycache = PickleCachePy(None, 10000)
>>> pyobj._p_jar = object()
>>> key = pyobj._p_oid = b'00000000'
>>> pycache[key] = pyobj
  File "//persistent/persistent/picklecache.py", line 154, in __setitem__
    self.data[oid] = value
  File "//lib/python3.7/weakref.py", line 168, in __setitem__
    self.data[key] = KeyedRef(value, self._remove, key)
  File "//lib/python3.7/weakref.py", line 337, in __new__
    self = ref.__new__(type, ob, callback)
TypeError: cannot create weak reference to 'PersistentPy' object
```

This creates a test failure in ZODB:

```console
$ PURE_PYTHON=1 zope-testrunner —test-path=src 
Failure in test doctest_invalidate (ZODB.tests.testConnection.InvalidationTests)
Failed doctest test for ZODB.tests.testConnection.InvalidationTests.doctest_invalidate
  File "/ZODB/src/ZODB/tests/testConnection.py", line 516, in doctest_invalidate

File "/ZODB/src/ZODB/tests/testConnection.py", line 532, in ZODB.tests.testConnection.InvalidationTests.doctest_invalidate
Failed example:
    transaction.commit()
Exception raised:
      File "//lib/python3.7/doctest.py", line 1329, in __run
        compileflags, 1), test.globs)
      File "<doctest ZODB.tests.testConnection.InvalidationTests.doctest_invalidate[9]>", line 1, in <module>
        transaction.commit()
      File "/transaction/transaction/_manager.py", line 257, in commit
        return self.manager.commit()
      File "/transaction/transaction/_manager.py", line 135, in commit
        return self.get().commit()
      File "/transaction/transaction/_transaction.py", line 282, in commit
        reraise(t, v, tb)
      File "/transaction/transaction/_compat.py", line 45, in reraise
        raise value
      File "/transaction/transaction/_transaction.py", line 273, in commit
        self._commitResources()
      File "/transaction/transaction/_transaction.py", line 465, in _commitResources
        reraise(t, v, tb)
      File "/transaction/transaction/_compat.py", line 45, in reraise
        raise value
      File "/transaction/transaction/_transaction.py", line 439, in _commitResources
        rm.commit(self)
      File "/ZODB/src/ZODB/Connection.py", line 497, in commit
        self._commit(transaction)
      File "/ZODB/src/ZODB/Connection.py", line 546, in _commit
        self._store_objects(ObjectWriter(obj), transaction)
      File "/ZODB/src/ZODB/Connection.py", line 609, in _store_objects
        self._cache[oid] = obj
      File "/persistent/persistent/picklecache.py", line 144, in __setitem__
        self.data[oid] = value
      File "///lib/python3.7/weakref.py", line 168, in __setitem__
        self.data[key] = KeyedRef(value, self._remove, key)
      File "///lib/python3.7/weakref.py", line 337, in __new__
        self = ref.__new__(type, ob, callback)
    TypeError: cannot create weak reference to 'Persistent' object
```

(Why don't we see this failure running ZODB with PyPy? I'm not sure, but I'd guess that we don't even try to run this test; for some reason, PyPy runs only 814 ZODB unit tests while CPython runs 1185. In total, PyPy runs 1313 tests while CPython runs 1684 tests.)

jamadden:
Except on PyPy, where we can already weakref them automatically. On CPython, track them using their `id` (can't use an `ffi.new_handle` "pointer" because it strongly references the object it's a handle for).

Fixes #133.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,12 @@
 - Add support for Python 3.9a3+.
   See `issue 124 <https://github.com/zopefoundation/persistent/issues/124>`_.
 
+- Fix the Python implementation of the PickleCache to be able to store
+  objects that cannot be weakly referenced. See `issue 133
+  <https://github.com/zopefoundation/persistent/issues/133>`_.
+
+  Note that ``ctypes`` is required to use the Python implementation
+  (except on PyPy).
 
 4.5.1 (2019-11-06)
 ------------------
--- a/persistent/picklecache.py
+++ b/persistent/picklecache.py
@@ -12,13 +12,14 @@
 #
 ##############################################################################
 import gc
-import weakref
 
+from weakref import WeakValueDictionary
 
 from zope.interface import implementer
 from zope.interface import classImplements
 
 from persistent._compat import use_c_impl
+from persistent._compat import PYPY
 from persistent.interfaces import GHOST
 from persistent.interfaces import IPickleCache
 from persistent.interfaces import IExtendedPickleCache
@@ -38,15 +39,10 @@ __all__ = [
 # pylint:disable=protected-access
 
 
-# On Jython, we need to explicitly ask it to monitor
-# objects if we want a more deterministic GC
-if hasattr(gc, 'monitorObject'): # pragma: no cover
-    _gc_monitor = gc.monitorObject # pylint:disable=no-member
-else:
-    def _gc_monitor(o):
-        pass
 
 _OGA = object.__getattribute__
+_OSA = object.__setattr__
+
 
 def _sweeping_ring(f):
     # A decorator for functions in the PickleCache
@@ -60,6 +56,82 @@ def _sweeping_ring(f):
         finally:
             self._is_sweeping_ring = False
     return locked
+
+
+class _WeakValueDictionary(object):
+    # Maps from OID -> Persistent object, but
+    # only weakly references the Persistent object. This is similar
+    # to ``weakref.WeakValueDictionary``, but is customized depending on the
+    # platform. On PyPy, all objects can cheaply use a WeakRef, so that's
+    # what we actually use. On CPython, though, ``PersistentPy`` cannot be weakly
+    # referenced, so we rely on the fact that the ``id()`` of an object is its
+    # memory location, and we use ``ctypes`` to cast that integer back to
+    # the object.
+    #
+    # To remove stale addresses, we rely on the ``ffi.gc()`` object with the exact
+    # same lifetime as the ``PersistentPy`` object. It calls us, we get the ``id``
+    # back out of the CData, and clean up.
+    if PYPY: # pragma: no cover
+        def __init__(self):
+            self._data = WeakValueDictionary()
+
+        def _from_addr(self, addr):
+            return addr
+
+        def _save_addr(self, oid, obj):
+            return obj
+
+        cleanup_hook = None
+    else:
+        def __init__(self):
+            # careful not to require ctypes at import time; most likely the
+            # C implementation is in use.
+            import ctypes
+
+            self._data = {}
+            self._addr_to_oid = {}
+            self._cast = ctypes.cast
+            self._py_object = ctypes.py_object
+
+        def _save_addr(self, oid, obj):
+            i = id(obj)
+            self._addr_to_oid[i] = oid
+            return i
+
+        def _from_addr(self, addr):
+            return self._cast(addr, self._py_object).value
+
+        def cleanup_hook(self, cdata):
+            oid = self._addr_to_oid.pop(cdata.pobj_id, None)
+            self._data.pop(oid, None)
+
+    def __contains__(self, oid):
+        return oid in self._data
+
+    def __len__(self):
+        return len(self._data)
+
+    def __setitem__(self, key, value):
+        addr = self._save_addr(key, value)
+        self._data[key] = addr
+
+    def pop(self, oid):
+        return self._from_addr(self._data.pop(oid))
+
+    def items(self):
+        from_addr = self._from_addr
+        for oid, addr in self._data.items():
+            yield oid, from_addr(addr)
+
+    def get(self, oid, default=None):
+        addr = self._data.get(oid, self)
+        if addr is self:
+            return default
+        return self._from_addr(addr)
+
+    def __getitem__(self, oid):
+        addr = self._data[oid]
+        return self._from_addr(addr)
 
 
 @use_c_impl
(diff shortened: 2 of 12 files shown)


persistent/picklecache.py
```diff
@@ -62,6 +65,87 @@ def locked(self, *args, **kwargs):
...
+
+        try:
+            from ctypes import cast
+            from ctypes import py_object
+            cast = staticmethod(cast)
+        except ImportError: # pragma: no cover
+            def _from_addr(self, addr):
+                for i in gc.get_objects():
+                    if id(i) == addr:
+                        return i
```
mgedmin: I don't like this.  id() values can be reused for unrelated objects in CPython.

persistent/picklecache.py
```diff
@@ -379,17 +463,14 @@ def _sweep(self, target, target_size_bytes=0):
...
-                    # Test-cases sneak in non-Persistent objects, sigh, so naturally
-                    # they don't cooperate (without this check a bunch of test_picklecache
-                    # breaks)
-                    or not isinstance(value, self._SWEEPABLE_TYPES)):
-                    to_eject.append((i, value))
+                        # Test-cases sneak in non-Persistent objects, sigh, so naturally
+                        # they don't cooperate (without this check a bunch of test_picklecache
+                        # breaks)
+                        or not isinstance(value, self._SWEEPABLE_TYPES)):
+                    ring.delete_node(node)
```
mgedmin: Just double-checking: does `self.ring` support mutation during iteration?

persistent/ring.py
```diff
@@ -108,57 +97,93 @@ class _CFFIRing(object):
...
     def iteritems(self):
         head = self.ring_home
         here = head.r_next
+        ring_to_obj = self.ring_to_obj
         while here != head:
-            yield here
+            # We allow mutation during iteration, which means
+            # we must get the next ``here`` value before
+            # yielding, just in case the current value is
+            # removed.
```
mgedmin: And this nicely answers my question ;)

persistent/picklecache.py
jamadden: Very true! 

I don't expect this to be used on CPython. Even if it were, the ref counting semantics that are tied to `ffi.gc` should eliminate the race condition.

persistent/picklecache.py
jamadden: Happily now it does!

I took this opportunity to improve the semantics of the one-ond-only Ring implementation. It resulted in, i think, some nice simplifications.

persistent/picklecache.py
jamadden: That fallback loop was borrowed from `objgraph`, IIRC. Perhaps the state of the art for non-CPython implementations has moved on since I checked?

jamadden:
> I'm very unqualified to review this (what even is ffi.gc? I probably don't even want to know), but other than that this looks very plausible to me. Except for the gc.get_objects() loop. 

The `gc.get_objects` loop is for, nominally, Jython support. Such support is still claimed and supported by comments in a few places. That once mattered a lot to me, but less so now, so if it's a problem I'm happy eliding it. 

The ZODB `PURE_PYTHON` tests on CPython run now with this change…with the exception of a block that's specifically targeting Jython, and one that expects ref counts of `None` to remain stable: yikes! (I think I might have written that.)

> what even is ffi.gc? I probably don't even want to know

Indeed. My first few attempts resulted in segfaults everywhere, and then in undefined behaviour. But I think (hope?) this is close to the simplest possible exploitation of it: a reliable ref-counted object that doesn't depend on the circular CPython GC or `__weakref__`, yet still delivers a callback. (My first tries used CFFI handles, which wound up using cycles, which is not a good practice. I squashed that history away.)

This doesn't have any size benefit over `__weakref__` on CPython, I think. It just has consistency benefits. Maybe that is or is not enough.

persistent/picklecache.py
mgedmin: objgraph is a debugging tool.  I allow hacks in debugging code that I wouldn't allow in production code ;)

mgedmin: I think my primary reason for rejecting this is that AFAIU `_from_addr()` is supposed to return None if the object associated with the old id() value has been freed, but the get_objects() method cannot reliably implement that.

icemac:
Is Jython still a thing and is anyone using it together with zopefoundation stuff? Currently we are not even testing (at least not automatically) on Jython so I feel not very well in claiming that we still support it.

jamadden:
> Is Jython still a thing and is anyone using it together with zopefoundation stuff? 

It's still a thing. As for its use with this stuff, I don't know anymore.

> Currently we are not even testing (at least not automatically) on Jython so I feel not very well in claiming that we still support it.

We don't really claim to support it (it's not in the classifiers, for example). There are just some remnants of support for it left in comments and alternate code paths.

However, as of persistent 4.5.0, which made CFFI a firm requirement, all possibility of using Jython was eliminated (until such time as the CFFI port is finished). (I should have remembered that.) So there's really no need for this code path and I can drop it.

CHANGES.rst
```diff
@@ -44,6 +44,9 @@
   objects that cannot be weakly referenced. See `issue 133
   <https://github.com/zopefoundation/persistent/issues/133>`_.
 
+  Note that ``cytypes`` is required to use the Python implementation
```
mgedmin: Is that a typo of `ctypes`?

jamadden: Thanks!

d-maurer:
I currently have the idea to use a `WeakKeyDictionayy` (with an
index object (i.e. a persistent object) as key) to
solve "https://github.com/zopefoundation/Products.ZCatalog/issues/94".
Does this become impossible with your optimizations?

jamadden:
I believe the two are unrelated. The changes here have nothing to do with making objects weakly referenceable or not. All that changed here is that the pure Python PickleCache can now store persistent objects that *cannot* be weakly referenced; the C implementation has always been able to do this. (All `Persistent` objects that come from ZODB are stored in a PickleCache.) By default, subclasses of `Persistent` *can* be weakly referenced.

PR #140, opened 2020-03-12
Remove faulty assertion in PersistentMapping.clear().

### LINKED ISSUE #139: AssertionError clearing PersistentMap not added to connection.
```pycon
>>> import persistent.mapping
>>> pm = persistent.mapping.PersistentMapping()
>>> pm['k'] = 'v'
>>> pm.clear()
  File "/Users/jmadden/Projects/GithubSources/persistent/persistent/mapping.py", line 104, in clear
    assert self._p_changed
AssertionError
```

This was a bug introduced in https://github.com/zopefoundation/persistent/pull/127 that slipped through review.

jamadden:
If there's no jar, we can't be _p_changed.

Fixes #139.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@
 4.6.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix an ``AssertionError`` clearing a non-empty ``PersistentMapping``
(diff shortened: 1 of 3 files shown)


jamadden:
Thanks!

Issue #137, opened 2020-03-06
Segfault in persistent 4.6.0


jamesjer:
I maintain the persistent package for Fedora.  I attempted to build version 4.6.0 for Fedora Rawhide.  The build was successful on all 64-bit architectures, but the test suite failed with segfaults on both 32-bit architectures: i686 and arm.  The segfaults are repeatable; I get them every time.  Running under valgrind shows no problems and the test suite completes successfully.  Running under gdb, however, I was able to gather some information.  The faulting test is `test_incrgc_simple (persistent.tests.test_picklecache.PythonPickleCacheTests)`.  The fault happens inside `PyObject_Repr`, on a bogus (garbage collected?) object.  This is what gdb has to say about the object:
```
(gdb) print *v
$1 = {ob_refcnt = -159554747, ob_type = 0x56703a34}
(gdb) print *$1->ob_type
$2 = {ob_base = {ob_base = {ob_refcnt = 0, ob_type = 0x0}, 
    ob_size = -158259432}, tp_name = 0x0, tp_basicsize = 0, tp_itemsize = 0, 
  tp_dealloc = 0x11, tp_vectorcall_offset = 1451167352, 
  tp_getattr = 0x566634d8, tp_setattr = 0x10, tp_as_async = 0x2e0, 
  tp_repr = 0x1, tp_as_number = 0xf7df84c0 <PyUnicode_Type>, 
  tp_as_sequence = 0x2be, tp_as_mapping = 0xffffffff, tp_hash = 0xa0d0de4, 
  tp_call = 0x0, tp_str = 0x70732041, tp_getattro = 0x61696365, 
  tp_setattro = 0x6f63206c, tp_as_buffer = 0x7274736e, tp_flags = 544498549, 
  tp_doc = 0x62617375 <error: Cannot access memory at address 0x62617375>, 
  tp_traverse = 0x7420656c, tp_clear = 0x6e61206f, 
  tp_richcompare = 0x61746f6e, tp_weaklistoffset = 1663067508, 
  tp_iter = 0x7373616c, tp_iternext = 0x6a626f20, tp_methods = 0x73746365, 
  tp_members = 0x200a0a2e, tp_getset = 0x46202020, tp_base = 0x6520726f, 
  tp_dict = <unknown at remote 0x706d6178>, tp_descr_get = 0x202c656c, 
  tp_descr_set = 0x70707573, tp_dictoffset = 543519599, tp_init = 0x68206577, 
[...]

PR #141, opened 2020-03-18
Use unsigned integers to represent addresses

jamesjer:
This fixes https://github.com/zopefoundation/persistent/issues/137.

### DIFF
--- a/persistent/_ring_build.py
+++ b/persistent/_ring_build.py
@@ -31,7 +31,7 @@ typedef struct CPersistentRingCFFI_struct
 {
     struct CPersistentRing_struct *r_prev;
     struct CPersistentRing_struct *r_next;
-    intptr_t pobj_id; /* The id(PersistentPy object) */
+    uintptr_t pobj_id; /* The id(PersistentPy object) */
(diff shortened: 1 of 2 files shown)


jamesjer:
I should mention that I tested this, and it does fix the problem for me on our 32-bit builders.

jamadden:
CPython's own `id()` builtin casts to `uintptr_t`:

https://github.com/python/cpython/blob/da1fe768e582387212201ab8737a1a5f26110664/Python/bltinmodule.c#L1168-L1172
```C
static PyObject *
builtin_id(PyModuleDef *self, PyObject *v)
{
    PyObject *id = PyLong_FromVoidPtr(v);
…
}
```

https://github.com/python/cpython/blob/da1fe768e582387212201ab8737a1a5f26110664/Objects/longobject.c#L1086-L1091
```C
PyObject *
PyLong_FromVoidPtr(void *p)
{
#if SIZEOF_VOID_P <= SIZEOF_LONG
    return PyLong_FromUnsignedLong((unsigned long)(uintptr_t)p);
#else
…
}
```

PR #145, opened 2020-03-26
Fix a test failure against zope.interface5 due to overly specific exception tests.

jamadden:
Fixes #144

PR #146, opened 2020-03-26
 Fix two reference leaks that could occur following an obscure error

### LINKED ISSUE #143: Two potential reference leaks
As was pointed out in my previous pull request, I haven't signed any contributor agreements.  Therefore, instead of submitting another pull request, I'm just going to point out two more near-trivial changes that should be made to the code.  Both apply to situations that probably never happen in practice, so this is most likely just a code cleanliness exercise.

In persistent/cPersistence.c, the `Py_DECREF(ts_module)` on line 1283 should be moved above the `if` statement on line 1281.  Otherwise, if `TimeStamp` is not found, a reference to `ts_module` is leaked.

In persistent/cPickleCache.c, the `if` statement on line 638 just returns if `PyDict_DelItem` returns a negative value.  But in that case, `dead_pers_obj` was not removed from the dictionary, and therefore its reference count was not decremented.  The body of the `if` statement should first `Py_DECREF(dead_pers_obj)` and only then return.  Actually, I'm not sure it should return.  Shouldn't the rest of the function be executed anyway?

jamadden:
Fixes #143

Based on #145 (to avoid the guaranteed conflict in CHANGES.rst), only the last commit is relevant.

### DIFF
--- a/persistent/cPickleCache.c
+++ b/persistent/cPickleCache.c
@@ -193,16 +193,16 @@ scan_gc_items(ccobject *self, int target, Py_ssize_t target_bytes)
     */
     insert_after(&before_original_home, self->ring_home.r_prev);
     here = self->ring_home.r_next;   /* least recently used object */
-	/* All objects should be deactivated when the objects count parameter
-	 * (target) is zero and the size limit parameter in bytes(target_bytes)
-	 * is also zero.
-	 *
-	 * Otherwise the objects should be collect while one of the following
-	 * conditions are True:
-	 *  - the ghost count is bigger than the number of objects limit(target).
-	 *  - the estimated size in bytes is bigger than the size limit in
-	 *    bytes(target_bytes).
-	 */
+    /* All objects should be deactivated when the objects count parameter
+     * (target) is zero and the size limit parameter in bytes(target_bytes)
+     * is also zero.
+     *
+     * Otherwise the objects should be collect while one of the following
+     * conditions are True:
+     *  - the ghost count is bigger than the number of objects limit(target).
+     *  - the estimated size in bytes is bigger than the size limit in
+     *    bytes(target_bytes).
+     */
     while (here != &before_original_home &&
             (
              (!target && !target_bytes) ||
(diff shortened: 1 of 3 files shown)


persistent/cPickleCache.c
```diff
@@ -636,17 +641,31 @@ cc_oid_unreferenced(ccobject *self, PyObject *oid)
...
+           calls that could cause a greenlet switch so the state of the
+           dictionary should not change). We still need to finish the cleanup.
+
+           Just write an unraisable error (like an exception from __del__,
+           because that's basically what this is).
+        */
+        PyErr_WriteUnraisable((PyObject*)dead_pers_obj);
+        PyErr_Clear();
+	/* Have the same side effect of clearing a ref count as the dict would have.*/
+	Py_DECREF(dead_pers_obj);
```
mgedmin: The indentation seems off here.

mgedmin:
The travis failure is probably unrelated:

```
+ auditwheel repair wheelhouse/persistent-4.6.4.dev0-cp27-cp27m-linux_x86_64.whl -w /io/wheelhouse/
  File "/usr/local/bin/auditwheel", line 8, in <module>
    sys.exit(main())
  File "/opt/_internal/cpython-3.7.7/lib/python3.7/site-packages/auditwheel/main.py", line 50, in main
    rval = args.func(args, p)
  File "/opt/_internal/cpython-3.7.7/lib/python3.7/site-packages/auditwheel/main_repair.py", line 49, in execute
    from .repair import repair_wheel
  File "/opt/_internal/cpython-3.7.7/lib/python3.7/site-packages/auditwheel/repair.py", line 14, in <module>
    from .wheeltools import InWheelCtx, add_platforms
  File "/opt/_internal/cpython-3.7.7/lib/python3.7/site-packages/auditwheel/wheeltools.py", line 17, in <module>
    from wheel.wheelfile import WHEEL_INFO_RE  # type: ignore
ModuleNotFoundError: No module named 'wheel.wheelfile'
The command "bash .manylinux.sh" exited with 1.
```

persistent/cPickleCache.c
jamadden: Thanks. That file has a mix of tabs and spaces for indentation and it confuses my editor.

jamadden:
> The travis failure is probably unrelated:
> 
> + auditwheel repair wheelhouse/persistent-4.6.4.dev0-cp27-cp27m-linux_x86_64.whl -w /io/wheelhouse/
>  File "/opt/_internal/cpython-3.7.7/lib/python3.7/site-packages/auditwheel/wheeltools.py", line 17, in <module>
    from wheel.wheelfile import WHEEL_INFO_RE  # type: ignore
> ModuleNotFoundError: No module named 'wheel.wheelfile'
> The command "bash .manylinux.sh" exited with 1.

Yes, that's a known issue since manylinux updated to auditwheel 3.1 6 days ago. https://github.com/pypa/manylinux/issues/512

jamadden:
Upstream issue has been fixed, build caches have been cleared, new images downloaded, and the build is now green.

Thanks!

PR #148, opened 2020-06-09
Add support for static properties

penn5:
They are only assignable on the class rather than instances of it.
They can be retrieved without deghostifying classes.
They do not trigger any hooks

This is the first step towards supporting asyncio

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -99,6 +99,9 @@ static PyObject * pickle_slotnames(PyTypeObject *cls);
 
 static PyObject * convert_name(PyObject *name);
 
+static PyObject * get_static(cPersistentObject *self, PyObject *attr);
+static PyObject * is_static(cPersistentObject *self, PyObject *attr);
+
 /* Load the state of the object, unghostifying it.  Upon success, return 1.
  * If an error occurred, re-ghostify the object and return -1.
  */
@@ -152,6 +155,7 @@ unghostify(cPersistentObject *self)
 /****************************************************************************/
 
 static PyTypeObject Pertype;
+static PyTypeObject Statictype;
 
 static void
 accessed(cPersistentObject *self)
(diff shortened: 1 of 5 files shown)


mgedmin:
I don't see any documentation about the new feature?

persistent/cPersistence.c
```diff
@@ -917,21 +921,73 @@ unghost_getattr(const char *s)
...
+    }
+}
+
+static PyObject *
+get_static(cPersistentObject *self, PyObject *attr)
+{
+    PyObject *clattr;
+    clattr = 0;
+    clattr = is_static(self, attr);
+    PyObject *value;
```
mgedmin, line 959: It looks like the version of MSVC++ used to build persistent for Python 2.7 on Windows in Appveyr doesn't support C99 variable declarations in the middle of a block:

```
    C:\Users\appveyor\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\cl.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -Ic:\python27\include -Ic:\python27\PC /Tcpersistent/cPersistence.c /Fobuild\temp.win32-2.7\Release\persistent/cPersistence.obj
    cPersistence.c
    persistent/cPersistence.c(959) : error C2275: 'PyObject' : illegal use of this type as an expression
            c:\python27\include\object.h(108) : see declaration of 'PyObject'
    persistent/cPersistence.c(959) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(962) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(962) : warning C4047: '=' : 'int' differs in levels of indirection from 'PyObject *'
    persistent/cPersistence.c(963) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(966) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(966) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(966) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(966) : warning C4047: 'function' : 'PyObject *' differs in levels of indirection from 'int'
    persistent/cPersistence.c(966) : warning C4024: 'function through pointer' : different types for formal and actual parameter 1
    persistent/cPersistence.c(966) : warning C4133: 'function' : incompatible types - from 'cPersistentObject *' to 'PyObject *'
    persistent/cPersistence.c(966) : warning C4047: '=' : 'int' differs in levels of indirection from 'PyObject *'
    persistent/cPersistence.c(968) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(970) : error C2065: 'value' : undeclared identifier
    persistent/cPersistence.c(970) : warning C4047: 'return' : 'PyObject *' differs in levels of indirection from 'int'
    error: command 'C:\\Users\\appveyor\\AppData\\Local\\Programs\\Common\\Microsoft\\Visual C++ for Python\\9.0\\VC\\Bin\\cl.exe' failed with exit status 2
```

jamadden:
If this is for the strawman outlined in https://github.com/zopefoundation/ZODB/issues/311, I think it would be prudent to see more of a full stack showing that this actually accomplishes something useful, rather than merging in isolation. If "asyncio support in ZODB" is a direction that's interesting, it will probably take more thought and design than what I sketched out there.

This also seems like it should be done by a new subclass of `Persistent`; it doesn't have to add overhead to every existing instance of `Persistent` (none of which are going to be asyncio-ready anyway).

penn5:
> If this is for the strawman outlined in [zopefoundation/ZODB#311](https://github.com/zopefoundation/ZODB/issues/311), I think it would be prudent to see more of a full stack showing that this actually accomplishes something useful, rather than merging in isolation. If "asyncio support in ZODB" is a direction that's interesting, it will probably take more thought and design than what I sketched out there.
> 
> This also seems like it should be done by a new subclass of `Persistent`; it doesn't have to add overhead to every existing instance of `Persistent` (none of which are going to be asyncio-ready anyway).

The idea is that Persistent doesn't actually perform any intensive operations, only the backend it uses (the jar). My plan is to make it passthrough the return value of the operations so it can return a coroutine, which the consumer then awaits. However setattr and delattr don't support this so I'm working on a dirty-ish hack

penn5:
> I don't see any documentation about the new feature?

¯\_(ツ)_/¯

icemac:
What is the status of this PR, @penn5 are you still planning to work on it?

penn5:
No.

PR #156, opened 2021-03-02
Ask zest.releaser not to upload binary wheels.

### LINKED ISSUE #154: Issue in macOS binary wheel for 4.6.4/3.8: Illegal Instruction (my fault, but needs a change to prevent in the future)
I'm working on moving RelStorage from Travis CI to GitHub actions. I got to the point of being able to import RelStorage, and [the process crashed importing `persistent`](https://github.com/zodb/relstorage/runs/1793041090?check_suite_focus=true#step:17:28) with an Illegal Instruction error.

```python traceback
Fatal Python error: Illegal instruction

Current thread 0x000000011119edc0 (most recent call first):
...
  File "//site-packages/persistent/_compat.py", line 81 in _c_optimizations_available
  File "//site-packages/persistent/_compat.py", line 182 in find_impl
  File "//site-packages/persistent/_compat.py", line 190 in use_c_impl
  File "//site-packages/persistent/interfaces.py", line 27 in <module>
...
  File "//site-packages/persistent/__init__.py", line 33 in <module>
...
  File "//relstorage/src/relstorage/adapters/adapter.py", line 26 in <module>
```

I've seen this previously with other binary wheels built on macOS when the binary was built on a newer computer than the destination and using `-march=native` or similar optimizations.

And it turns out that there are two wheels for Python 3.8 published on PyPI for 4.6.4, one that `zest.releaser` built and uploaded for me (cp38-10.15), and one that came from Travis (cp38-10.9). My environment includes `-march=native` in `CFLAGS` et al.

I will yank the bad wheel, solving the immediate problem.

In other projects (gevent), I've used a zest.releaser hook to check the environment for that setting and forbid uploading a binary wheel built that way.

But IIRC zest.releaser now also has a way to say "just upload the source, never a binary", and that might be the better thing to do here, if we can continue relying on CI to build wheels.

jamadden:
We use the CI to build and upload binary wheels.

Binary wheels built on individuals' machines may use CFLAGS, etc, that are not compatible with all machines, but this is not expressed in the wheel tags. Thus, incompatible machines may download and install such wheels, leading to bad behaviour or even Illegal Instruction crashes.

Fixes #154.

Also set the long_description_content_type to avoid getting warnings from 'twine check'.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,12 @@
 [aliases]
 dev = develop easy_install persistent[testing]
 docs = develop easy_install persistent[docs]
+
+[zest.releaser]
+create-wheel = no
+
+[bdist_wheel]
+universal = 0
+
+[metadata]
+long_description_content_type = text/x-rst


icemac (approved):
LGMT.

Do we have to make this change in every repository where we have C code involved?

mgedmin:
> Do we have to make this change in every repository where we have C code involved?

I'd say yes.  Luckily there aren't that many of them.

mgedmin (approved):

jamadden:
Thanks!

PR #160, opened 2021-03-10
Make the Python impl handle setting __class__ and __dict__ on ghosts like C

### LINKED ISSUE #155: C/Python difference: Setting __class__ activates in C, doesn't activate in Python
In the C implementation, setting `__class__` of a ghost calls `_p_activate()`, and uses the class methods defined on the original class. 

But in the Python implementation, setting `__class__` does *not* call `_p_activate()`; consequently, the class methods of the *new* class are used. 

This makes a difference when class methods include those used in the pickle protocol, such as `__setstate__`.

Here's an example. Consider this code:
```python
from ZODB.DB import DB
from ZODB.DemoStorage import DemoStorage
import transaction

class A(Persistent):

    def __setstate__(self, state):
        print("Setting state in A")
        Persistent.__setstate__(self, state)

class B(Persistent):

    def __setstate__(self, state):
        print("Setting state in B")
        Persistent.__setstate__(self, state)


db = DB(DemoStorage())
transaction.begin()
conn = db.open()
conn.root()['key'] = A()
transaction.commit()

transaction.begin()
conn2 = db.open()
a = conn2.root()['key']
print("Setting the class")
a.__class__ = B
print("Activating the object")
a._p_activate()
transaction.abort()
```

When we run it with the C implementation, the object is activated by the line `a.__class__ = B` and uses the `A.__setstate__` implementation to do so.

```console
$ python /tmp/t.py
Setting the class
Setting state in A
Activating the object
Done activating
```

Using the pure-Python version, however, activation doesn't happen until we call `_p_activate()` explicitly, and then it uses the `Other.__setstate__` implementation.

```console
PURE_PYTHON=1 python /tmp/t.py
Setting the class
Activating the object
Setting state in B
Done activating
```

Ideally, setting the class of a ghost wouldn't activate the object, I think. But since that's the way the C implementation has worked for some time, probably the Python implementation needs to change to activate the object when the class is assigned to.

jamadden:
By activating the object.

Add tests for this.

In the process of adding tests, I discovered the mock pickle cache used by the Python tests was returning the wrong thing from new_ghost (not a ghost), so I fixed this and added some extra tests and assertions to be sure the behaviour matches with the real and C caches.

Fixes #155

### DIFF
--- a/persistent/cPersistence.c
+++ b/persistent/cPersistence.c
@@ -973,7 +973,9 @@ Done:
 }
 
 /*
-  TODO:  we should probably not allow assignment of __class__ and __dict__.
+ * TODO:  we should probably not allow assignment of __class__ and __dict__.
+ * But we do, and have for a long time. Code in the wild is known to rely on
+ * assigning __class__.
 */
 
 static int
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -2041,18 +2041,14 @@ class PyPersistentTests(unittest.TestCase, _Persistent_Base):
         return PickleCache(jar, 10)
 
     def _checkMRU(self, jar, value):
-        try:
-            self.assertEqual(list(jar._cache._mru), value)
-        except AttributeError:
+        if not isinstance(jar._cache, self._getRealCacheClass()):
             # We can't do this for the real cache.
-            assert isinstance(jar._cache, self._getRealCacheClass())
+            self.assertEqual(list(jar._cache._mru), value)
 
     def _clearMRU(self, jar):
-        try:
-            jar._cache._mru[:] = []
-        except AttributeError:
+        if not isinstance(jar._cache, self._getRealCacheClass()):
             # We can't do this for the real cache.
-            assert isinstance(jar._cache, self._getRealCacheClass())
+            jar._cache._mru[:] = []
 
     def test_accessed_with_jar_and_oid_but_not_in_cache(self):
         # This scenario arises in ZODB: ZODB.serialize.ObjectWriter
--- a/persistent/tests/test_persistence.py
+++ b/persistent/tests/test_persistence.py
@@ -50,7 +50,7 @@ class _Persistent_Base(object):
     def _makeOne(self, *args, **kw):
         return self._getTargetClass()(*args, **kw)
 
-    def _makeJar(self):
+    def _makeJar(self, real_cache=False):
         from zope.interface import implementer
         from persistent.interfaces import IPersistentDataManager
 
@@ -77,7 +77,7 @@ class _Persistent_Base(object):
                 self._registered.append(obj._p_oid)
 
         jar = _Jar()
-        jar._cache = self._makeCache(jar)
+        jar._cache = self._makeRealCache(jar) if real_cache else self._makeCache(jar)
         return jar
 
     def _makeBrokenJar(self):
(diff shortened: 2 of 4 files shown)


persistent/tests/test_persistence.py
```diff
@@ -1968,15 +2032,27 @@ def update_object_size_estimation(self, oid, new_size):
...
+        PickleCache = self._getRealCacheClass()
         return PickleCache(jar, 10)
 
     def _checkMRU(self, jar, value):
-        self.assertEqual(list(jar._cache._mru), value)
+        try:
+            self.assertEqual(list(jar._cache._mru), value)
+        except AttributeError:
+            # We can't do this for the real cache.
+            assert isinstance(jar._cache, self._getRealCacheClass())
```
mgedmin, line 2048: Wouldn't this be clearer?

```suggestion
        if not isinstance(jar._cache, self._getRealCacheClass()):
            # We can't do this for the real cache.
            self.assertEqual(list(jar._cache._mru), value)
```

persistent/tests/test_persistence.py
```diff
@@ -1968,15 +2032,27 @@ def update_object_size_estimation(self, oid, new_size):
...
 
     def _clearMRU(self, jar):
-        jar._cache._mru[:] = []
+        try:
+            jar._cache._mru[:] = []
+        except AttributeError:
+            # We can't do this for the real cache.
+            assert isinstance(jar._cache, self._getRealCacheClass())
```
mgedmin, line 2055: Likewise 
```suggestion
        if not isinstance(jar._cache, self._getRealCacheClass()):
            # We can't do this for the real cache.
            jar._cache._mru[:] = []
```
?

mgedmin (approved):
LGTM

persistent/cPersistence.c
```diff
@@ -973,7 +973,9 @@ Per__p_getattr(cPersistentObject *self, PyObject *name)
 }
 
 /*
-  TODO:  we should probably not allow assignment of __class__ and __dict__.
+ * TODO:  we should probably not allow assignment of __class__ and __dict__.
+ * But we do, and have for a long time. Code in the wild is known to rely on
+ * assigning __class__.
```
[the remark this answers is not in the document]
jamadden: Sure. One example is these [two](https://github.com/NextThought/nti.site/blob/bf6980eb8ee63b8c12979d498fec1b1dc7c7f3fe/src/nti/site/site.py#L392-L398) related [parts]( https://github.com/NextThought/nti.site/blob/bf6980eb8ee63b8c12979d498fec1b1dc7c7f3fe/src/nti/site/site.py#L441-L448) of `nti.site`.

Basically, the `Components` object of `zope.interface` implements `__bases__` by having its tree of objects copy references to the two adapter registries that make up a `Components`. This means that if we need to change the implementation of one of those adapter registries, we can't do so by just assigning to, e.g., `self.adapters`: there are likely many other references to the old object still around somewhere.  Thus if we want to make a change, the object has to be mutated *in place*.  For behaviour changes in methods, that means assigning to `__class__`.

This problem is magnified by persistence, especially because, while the adapter registries from `zope.component.persistentregistery` are `Persistent`, the `PersistentComponents` actually is *not*, making migration painful.

This is ultimately the motivation for https://github.com/zopefoundation/zope.interface/pull/228

persistent/tests/test_persistence.py
jamadden: I like the `try/except` version because it makes two assertions: the mock object works one way, and if that fails, then we must be using the real object. The `if` version kind of implicitly puts that second assertion in the condition. It seemed to me that the former is "stronger" and more likely to break if something unexpectedly changes.

But that's really all just a gut feeling. I don't mind changing it.

jamadden:
Thank you!

PR #161, opened 2021-03-10
Fix the CFFI cache implementation to not print AttributeErrors.

### LINKED ISSUE #150: AttributeError: '_WeakValueDictionary' object has no attribute '_addr_to_oid'
Seen during test teardown of a large app using CPython in PURE_PYTHON=1 mode

```
From callback for ffi.gc <cdata 'struct CPersistentRingCFFI_struct *' owning 24 bytes>:
  File "//persistent/persistent/picklecache.py", line 105, in cleanup_hook
    oid = self._addr_to_oid.pop(cdata.pobj_id, None)
AttributeError: '_WeakValueDictionary' object has no attribute '_addr_to_oid'
```

This is ignored, and the dict is already gone, so there shouldn't be any harm, but it is annoying.

jamadden:
Only affects CPython when PURE_PYTHON=1.

Add tests for this. They're a bit dicey because this is a hard situation to get into, but they do test the symptoms.

Fixes #150

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@
   when setting its ``__class__`` and ``__dict__``. This matches the
   behaviour of the C implementation. See `issue 155
   <https://github.com/zopefoundation/persistent/issues/155>`_.
+- Fix the CFFI cache implementation (used on CPython when
+  ``PURE_PYTHON=1``) to not print unraisable ``AttributeErrors`` from
+  ``_WeakValueDictionary`` during garbage collection. See `issue 150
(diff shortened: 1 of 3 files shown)


jamadden:
I think this is a pretty simple change. 

Are there any interested reviewers?

icemac (approved):
The changes look sane to me.
I tested by raising the attribute error in `cleanup_hook`, it lead to a printed AttributeError (probably from `test_interpreter_finalization_ffi_cleanup`) and an failure from `test_cleanup_hook_gc`.

jamadden:
Thank you!

PR #162, opened 2021-03-10
Make the Python PickleCache run a GC when it evicts weakly-referenced objects.

### LINKED ISSUE #149: PyPy: Sweeping the cache may need to call gc.collect()
If there are weak references to objects that are ejected from the cache, those won't immediately get set to None unless gc.collect() is called. This can manifest in a variety of ways but tends to lead to a `ConnectionStateError`. One way that's easy to reproduce this is with a persistent component registry. Given this test:
```python
from zope import interface

class IExtra(interface.Interface): # pylint:disable=inherit-non-class
    "An extra interface"

class IExtra2(interface.Interface): # pylint:disable=inherit-non-class
    "Another extra interface"

class Object(object):
    def __init__(self, *args):
        pass

class TestResetDbCaches(unittest.TestCase):
    # For complete reliability, you probably want to disable automatic gc in setUp()
    def test_persistent_site_closed(self):
        from zope.site.site import LocalSiteManager
        from ZODB import DB
        from ZODB.DemoStorage import DemoStorage
        import transaction

        interface.classImplements(LocalSiteManager, IExtra)

        db = DB(DemoStorage())
        transaction.begin()
        conn = db.open()
        site_man = LocalSiteManager(None)
        conn.root()[0] = site_man
        site_man.registerAdapter(Object, [IExtra], IExtra2)
        site_man.getAdapter(LocalSiteManager(None), IExtra2)
        transaction.commit()
        conn.close()

        conn.cacheMinimize()
        IExtra.changed(None) # This raises
```

You get this error:
```python traceback
Error in test test_persistent_site_closed (nti.testing.tests.test_zodb.TestResetDbCaches)
  File "//lib-python/2.7/unittest/case.py", line 329, in run
    testMethod()
  File "//test_zodb.py", line 319, in test_persistent_site_closed
    IExtra.changed(None) 
  File "//zope/interface/interface.py", line 498, in changed
    dependent.changed(originally_changed)
  File "//zope/interface/declarations.py", line 273, in changed
    return super(Implements, self).changed(originally_changed)
  File "//zope/interface/interface.py", line 498, in changed
    dependent.changed(originally_changed)
  File "//zope/interface/adapter.py", line 474, in changed
    super(AdapterLookupBase, self).changed(None)
  File "//zope/interface/adapter.py", line 444, in changed
    self._verify_ro = self._registry.ro[1:]
  File "//persistent/persistence.py", line 284, in __getattribute__
    oga(self, '_p_activate')()
  File "//persistent/persistence.py", line 407, in _p_activate
    jar.setstate(self)
  File "//ZODB/Connection.py", line 785, in setstate
    raise ConnectionStateError(msg)
ConnectionStateError: Shouldn't load state for zope.site.site._LocalAdapterRegistry 0x184a8d02c5ee0a2b when the connection is closed
```

The `_v_lookup` object of the `site_man`'s `adapter` was registered as a dependent of `<implementedBy zope.site.site.LocalSiteManager>`  by `site_man.getAdapter(LocalSiteManager(None), IExtra2)`, which specifically subscribes to it. That same `<implementedBy>` object is also registered with `IExtra`, so when `IExtra` is changed, the `_v_lookup` is given a notification…but the backing connection has been closed.

A `gc.collect()` at the end of sweeping the cache clears out the weak reference in the dependents, so the `_v_lookup` isn't called anymore and the problem goes away.

(This shows up with `IAttributeAnnotatable` if tests load the configuration of `zope.site` more than once, because the configure.zcml says that `LocalSiteManager` provides that interface.)

In this specific case, overriding `_p_deactivate` in `_LocalAdapterRegistry` to clear out any registration also solves the problem:

```python
class _LocalAdapterRegistry(
        zope.component.persistentregistry.PersistentAdapterRegistry,
        zope.location.Location,
):
    def _p_deactivate(self):
        if self._v_lookup is not None:
           self._v_lookup.changed(None)
        super(_LocalAdapterRegistry, self)._p_deactivate()
```

But (a) it was tricky to figure out exactly what object to call `changed()` on, and if you get it wrong the errors aren't helpful and (b) it seems like this is a general issue for any persistent objects that might have weak references to it, so solving it when sweeping the ring should solve a bunch of problems at once.

Proposed patch would look something like this:
```diff
--------------------------------------
modified: persistent/picklecache.py
--------------------------------------
@@ -424,6 +424,7 @@ class PickleCache(object):
     @_sweeping_ring
     def _sweep(self, target, target_size_bytes=0):
         ejected = 0
+        found_weak_refs = False
         ring = self.ring
         for node, value in ring.iteritems():
             if ((target or target_size_bytes) # pylint:disable=too-many-boolean-expressions
@@ -447,7 +448,10 @@ class PickleCache(object):
                 # we eject objects. Also, note that we can only take any of these
                 # actions if our _p_deactivate ran, in case of buggy subclasses.
                 # see _persistent_deactivate_ran
-
+                if not found_weak_refs:
+                    found_weak_refs |= getattr(value, '__weakref__', None) is not None
                 value._p_deactivate()
                 if (self._persistent_deactivate_ran
                         # Test-cases sneak in non-Persistent objects, sigh, so naturally
@@ -457,7 +459,8 @@ class PickleCache(object):
                     ring.delete_node(node)
                     ejected += 1
                     self.non_ghost_count -= 1
-
+        if ejected and found_weak_refs:
+            gc.collect()
         return ejected
```

jamadden:
This fixes hard-to-debug ZODB ConnectionStateError cases with persistent zope.component registrations on PyPy.

I tried to come up with a test that would also fail on CPython using circular references, but did not succeed (because part of evicting objects is clearing their dict, breaking the reference cycle).

Fixes #149. Well, band-aids it anyway. I'd like to think of a better way (the assumption that weakrefs are cleared immediately is pretty baked-in to zope.interface, at least when used with ZODB).

Currently based on #161 to avoid conflicts; only the last commit is relevant.

### DIFF
--- a/persistent/tests/test_picklecache.py
+++ b/persistent/tests/test_picklecache.py
@@ -44,6 +44,16 @@ class DummyConnection(object):
     pass
 
 
+class ClosedConnection(DummyConnection):
+    def __init__(self, test):
+        self.test = test
+
+    def setstate(self, obj): # pragma: no cover
+        self.test.fail("Connection is closed")
+
+    def register(self, obj):
+        """Does nothing."""
+
 def _len(seq):
     return len(list(seq))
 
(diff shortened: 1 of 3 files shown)


jamadden:
A version of this fix is present [in nti.testing](https://github.com/NextThought/nti.testing/pull/21) where it was necessary to test persistent zope.component registries under PyPy.

Are there any interested reviewers?

persistent/tests/test_picklecache.py
```diff
@@ -381,6 +391,84 @@ def test_full_sweep(self):
...
+        from weakref import ref as WeakRef
+        gc.disable()
+        self.addCleanup(gc.enable)
+
+        jar = ClosedConnection(self)
+        cache = self._makeOne(jar, 0)
+
+        # Make a persistent object, put it in the cache as saved
+        class P(self._getRealPersistentClass()):
+            "A real persistent object that can be weak referenced"
```
icemac, line 449:
```suggestion
            """A real persistent object that can be weak referenced."""
```

icemac (commented):
I cannot judge the code, as I do not know enough about it. 
I read the changes in the last commit and they look fine for me.
I tested locally by removing the `gc.collect()` call and it fails four tests on PyPy2 and PyPy3 but on CPython the call is not needed. That's fine as it was the goal of the PR.

jamadden:
> I read the changes in the last commit and they look fine for me. I tested locally…

Thank you for reviewing, and thank you for testing. I plan to move ahead with merging and releasing this now.

PR #168, opened 2022-02-01
Switch package to src-layout.

### LINKED ISSUE #158: Move to the src/ layout?
This is one of the few zopefoundation repos (mostly older ones) that use a bare directory in the root of the project instead of nesting it beneath src/. That makes it annoyingly semi-unique, and has some issues of its own.

Can we move persistent to src/persistent?

icemac:
Fixes #158.

This PR is the preparation to use `meta/config` later on and to support Python 3.11 in the end.


CHANGES.rst
```diff
@@ -5,7 +5,7 @@
 4.7.1 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Switch package to src-layout.
```
jamadden, line 8: I wonder if this needs a change note? It's not visible to end users. Even other packagers shouldn't care unless they were doing unspeakable things (i.e., not using 'pip wheel' or 'python setup.py bdist_wheel'). 

Not that adding the change note hurts, I just worry about confusing end users ("Do I need to upgrade? I don't even know what that means.") Maybe a sentence to the effect that "this is a packaging only change" and a reference to the PR or issue would help.

jamadden (approved):
LGTM and a big 👍 Thanks!

ETA: [The one appveyor failure](https://ci.appveyor.com/project/mgedmin/persistent/builds/42408028/job/14icedhj06k77scy#L76) was an unrelated network issue:
```
Found artifact 'dist\persistent-4.7.1.dev0-cp38-cp38-win_amd64.whl' matching 'dist\*.whl' path
Uploading artifacts...
Error uploading artifact to the storage: The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.
```

CHANGES.rst
icemac: I incorporated your feedback into https://github.com/zopefoundation/persistent/pull/168/commits/ea09894c0b3ee6774459c950c68eb42849103102.

icemac:

PR #170, opened 2022-03-09
Add support for Python 3.11 (as of 3.11a6).

icemac:


d-maurer (approved):

jamadden (approved):
LGTM, except I don't see a classifier added to `setup.py` for 3.11. Are we holding off for 3.11 final to do that?

icemac:

PR #177, opened 2022-09-15
Trigger GHA after deleting its caches.

icemac:
The new test checking against the `-Ofast` problem was failing on `master`.

I cleaned the caches using

```
repo=persistent;gh api -H "Accept: application/vnd.github+json" /repos/zopefoundation/$repo/actions/caches|jq -r .actions_caches[].key|sort -u|grep Linux-pip-|xargs -I{} gh api -H "Accept: application/vnd.github+json" --method DELETE /repos/zopefoundation/$repo/actions/caches?key={}
```

Rerunning the failed jobs on `master` was not enough, so I am doing a complete re-run here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,7 @@
 4.9.1 (unreleased)
 ==================
 
-- Update Python 3.11 support to 3.11.0-beta3.
+- Update Python 3.11 support to 3.11.0-rc1.
 
 - Disable unsafe math optimizations in C code.  See `pull request 176
   <https://github.com/zopefoundation/persistent/pull/176>`_.


icemac:
Hm, the same tests break as on `master`, @mgedmin do you have any idea how it was caused?

mgedmin:
Weird, I wiped the caches, re-ran failed jobs, and they succeeded.

The command I used looks identical to the one you used:

```
repo=persistent;gh api -H "Accept: application/vnd.github+json" /repos/zopefoundation/$repo/actions/caches|jq -r .actions_caches[].key|sort -u|grep Linux-pip-|xargs -I{} gh api -H "Accept: application/vnd.github+json" --method DELETE /repos/zopefoundation/$repo/actions/caches?key={}
```

mgedmin:
The git master build succeeded for me as well.

If this happens again, then maybe instead of playing whack-a-mole with the caches we ought to introduce a version in the cache key, e.g. Linux-v2-pip-pyX.Y? (v2 in front of pip so it doesn't match the old catch-all Linux-pip- prefix)

icemac:
@mgedmin That's strange, but it seems that it have worked. Thank you!

PR #180, opened 2022-12-20
Drop support for Python 2.7, 3.5, 3.6.

icemac:


dataflake (approved):

src/persistent/mapping.py
```diff
@@ -114,18 +109,15 @@ def update(self, *args, **kwargs):
         self.__super_update(*args, **kwargs)
         self._p_changed = 1
 
-    def setdefault(self, key, *args, **kwargs): # pylint:disable=arguments-differ
-        # (Python 3 and Python 2 have very different signatures.)
-
+    def setdefault(self, key, *args, **kwargs):
```
d-maurer, line 112: The signature could likely be simplified (same for other methods).

d-maurer (approved):

icemac:

PR #193, opened 2024-03-28
Flake8 the code.

icemac:
After fixing isort it got visible that we did not do this, yet.


dataflake (approved):
Do you want to remove the remaining pylint comments as well?

icemac:
> Do you want to remove the remaining pylint comments as well?

We probably could do this, but I am too lazy to do it now.

PR #198, opened 2024-05-24
Update meta/config template to publish Windows wheels to PyPI

dataflake:
Also see https://github.com/zopefoundation/meta/issues/235

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -17,6 +17,7 @@ source =
 [report]
 show_missing = true
 precision = 2
+ignore_errors = True
 exclude_lines =
     except ImportError:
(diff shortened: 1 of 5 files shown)


d-maurer (approved):
Do we need individual reviews for application of a "meta/config" template?
@tseaver merge without review with the justification "I only applied the template and all tests pass". I find this okay.

dataflake:
> Do we need individual reviews for application of a "meta/config" template? @tseaver merge without review with the justification "I only applied the template and all tests pass". I find this okay.

I agree and won't ask for reviews anymore if the change is small and the tests all pass. I was just following @icemac's habit.

PR #208, opened 2024-09-04
fix: remove 'simple_new' fossil

### LINKED ISSUE #206: Who uses `persistent.cPersistence.simple_new`?
I can't find any use of it in `BTrees` or `ZODB`.  AFAICT, it is a fossil, and it has some bad behaviour (see https://github.com/zopefoundation/persistent/pull/204/commits/da8067d2b4dfac5877d1cf2450f33f09097a2a70 which drops testing it with `type` and `tuple` because deallocating the returned instances trigger assertions).

I'm guessing it is left over from some ancient days:  I cannot find another ZODB-related project which uses it, nor any in the Zope3 and Zope2 repos I have hand.

tseaver:
Closes #206.

### DIFF
--- a/src/persistent/cPersistence.c
+++ b/src/persistent/cPersistence.c
@@ -28,9 +28,6 @@ struct ccobject_head_struct
 #include <stdint.h>
 #include <inttypes.h>
 
-/* These objects are initialized when the module is loaded */
-static PyObject *py_simple_new;
-
 /* Strings initialized by init_strings() below. */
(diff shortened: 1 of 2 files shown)


davisagli (approved):

src/persistent/cPersistence.c
icemac, line 1: Should we also have a change log entry mentioning the removal of this code?

tseaver: @icemac Done in 772493743c53972bbc4d58ff3341a00f1951b74d.

PR #214, opened 2025-02-24
Use Py_REFCNT to access cPersistentObject refcounts

cjwatson:
These assertions failed to compile if built without `-DNDEBUG`.  As of setuptools 75.7.0, setting `CFLAGS` to anything while building an extension no longer copies the `CFLAGS` that Python itself was built with, so `NDEBUG` is less likely to be defined.  See https://github.com/pypa/setuptools/issues/4836 and https://bugs.debian.org/1098602.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,13 +5,16 @@
 6.2 (unreleased)
 ================
 
+- Use ``Py_REFCNT`` to access ``cPersistentObject`` reference counts in
+  assertions.
+
 
 6.1 (2024-09-17)
 ================
 
 - Add final support for Python 3.13.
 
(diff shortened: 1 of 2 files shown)


tseaver (approved):
LGTM, although I don't really follow the pypa discussion.  In any case, using the `Py_REFCNT` macro[1] is cleaner than direct access.

[1] https://docs.python.org/3/c-api/refcounting.html#c.Py_REFCNT

thesamesam:
The TL;DR is that the asserts couldn't actually be compiled with some Python versions and setuptools was masking that by adding `-DNDEBUG` in more cases than one would expect.

icemac:
Just released in https://pypi.org/project/persistent/6.1.1/.

Issue #218, opened 2025-03-04
Unit tests fail on Python 3.14


dataflake:
Building the package under Python 3.14.0a5 and running the tests yields the following failures:

```python
Failure in test test__p_deactivate_from_saved_w_dict (persistent.tests.test_persistence.CPersistentTests.test__p_deactivate_from_saved_w_dict)
Traceback (most recent call last):
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 660, in run
    self._callTestMethod(testMethod)
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 606, in _callTestMethod
    result = method()
  File "/Users/jens/src/zope/persistent/.tox/py314/lib/python3.14/site-packages/persistent/tests/test_persistence.py", line 1343, in test__p_deactivate_from_saved_w_dict
    self.assertEqual(inst.__dict__, {})
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 916, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 1215, in assertDictEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 741, in fail
    raise self.failureException(msg)
AssertionError: {'normal': 'after'} != {}
- {'normal': 'after'}
+ {}

..........

Failure in test test__p_invalidate_from_changed_w_dict (persistent.tests.test_persistence.CPersistentTests.test__p_invalidate_from_changed_w_dict)
Traceback (most recent call last):
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 660, in run
    self._callTestMethod(testMethod)
  File "/Users/jens/src/python/Python-3.14.0a5/lib/python3.14/unittest/case.py", line 606, in _callTestMethod
    result = method()
[...]

PR #221, opened 2025-06-08
Clear dict py314

mamico:
Fix: https://github.com/zopefoundation/persistent/issues/218

IMHO in python 3.14 dictionary and/or garbage optimization requires to explicitly clear the dictionary.

Although this change seems to fix the tests in python 3.14 (tested with python 3.14.0b2), I suggest double-checking the change with someone more familiar with the internals of the library to exclude any side effects related to the change

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -102,6 +102,7 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14"
         os: [ubuntu-latest, macos-latest, windows-latest]
         exclude:
           - os: macos-latest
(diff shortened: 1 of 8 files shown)


dataflake (approved):
The setuptools issue is a separate problem, the dictionary fix should go ahead. I can't judge if that dictionary clearing has unwanted side effects, though.

davisagli:
Before this change, you can assign obj.__dict__ to another variable, then call obj._p_deactivate(), and still access the dict items via the other variable. After the change, the dict will be empty even if it's accessed via another reference. I'm not sure if that will cause problems but it seems worth noting.

davisagli (approved):

PR #225, opened 2025-11-10
Update Python version support.

icemac:
- **Update supported Python versions.**
- **Drop support for Python 3.9.**
- **Add support for Python 3.14.**
- **Configuring for c-code**

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -98,6 +98,7 @@ jobs:
           - "3.12"
           - "3.13"
           - "3.14"
+          - "3.15"
         os: [ubuntu-latest, macos-latest, windows-latest]
         exclude:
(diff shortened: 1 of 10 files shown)


icemac:
Might have the same problem described in https://github.com/zopefoundation/zope.interface/issues/348, so let's first find a solution there.

icemac:
> This can be merged safely. The problem with arm64 wheels only occurs when a tag is set and wheels are published.

This would be the next step I planned to do after merging the PR.

dataflake:
Updated the PR by applying the latest zope.meta templates, which fixes the arm64 wheel issue.

I also moved the metadata into ``pyproject.toml``, changed the docs theme to furo and cleaned up change log formatting.

